Issue #3, opened 2018-10-24
Proposal: Switch module narrative to application narrative


gamarin2:
Currently we focus on the module first, and then the application. It is true that the module is the main piece that people need to work on, but I think it's important for them to discover first how the app import and handle modules before coding one.

The bulk of the tuto would **still be the module**, but the narrative would change a bit. Namely, we would have:

- sdk-module-tutorial -> sdk-app-tutorial
- Instead of starting directly with keeper, we would start by an empty `./app.go` and then go in the module section when we realise we need it in the process of building app.go
- This would enable devs to have a more gentle introduction to how they can build their state machine and how the sdk manages modules.

Overall it would require little change, and would solidify this tutorial as a coherent, bottom-up and self-sufficient way of learning the sdk.

PR #6, opened 2018-10-24
Transition from module-focused to application-focused tutorial

jackzampolin:
Fixes: #3 

There are also some grammar updates here like added commas, and some changes to ensure consistent voice throughout the tutorial.

cc @ga


jackzampolin:
Ok, also ran through doing the tutorial and found a couple of bugs introduced by some recent changes to `develop`. I've pushed changes for that as well.

PR #9, opened 2018-11-09
Add send command to cli.

yangyanqing:
After adding command `nameservicecli tx send`, users can operate with multiple roles.


yangyanqing:
Need to improve it readme about multi roles

PR #10, opened 2018-11-09
Add send command to cli

yangyanqing:
After adding command `nameservicecli tx send`, users can operate with multiple roles.
Add specification for buying name from another, and format bash shell.


jackzampolin:
Great addition @yangyanqing!

yangyanqing:
Thanks for merging this PR. I want to add some testcase to this tutorial, how do you think about it? @jackzampolin

PR #12, opened 2018-11-19
move readme, for simplicity

zramsay:
this is where it'll be in the docs, so have it here.


gamarin2:
Made some adjustments in this other PR https://github.com/cosmos/sdk-application-tutorial/pull/14

PR #22, opened 2018-12-11
In tutorial, instruct to run `dep ensure` first

aknuds1:
In the tutorial, instruct the user to run `dep ensure -v` in order to bootstrap the dependency configuration as `dep init` will simply bail due to the fact that Gopkg.toml already exists.


jackzampolin:
Thank you @aknuds1, been meaning to fix this!

PR #23, opened 2018-12-18
fixed nsd's source code in the markdown 

MasashiSalvador57f:
Hi I'm going through this great tutorial, and I found that some part of the source codes in the markdowns don't seem like the same as the actual source code under the `x/nameservice` directory.

It might be helpful for the beginners if we can fix the tutorial markdown files. 😄 

### What I fixed 
- some lacked imports and some missing definitions of variable
- according to https://github.com/cosmos/sdk-application-tutorial/blob/master/cmd/nsd/main.go

### DIFF
--- a/tutorial/entrypoint.md
+++ b/tutorial/entrypoint.md
@@ -23,26 +23,31 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"io/ioutil"
 	"os"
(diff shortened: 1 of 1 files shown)


jackzampolin:
You are absolutely right! Thank you for the fix @MasashiSalvador57f!

PR #29, opened 2019-01-28
WIP: Billy/all names

okwme:
added a keystore for list of all names
probably not the most efficient way to do this
would love to know maybe how to use the iterator to do this or if there's another method?
anyway for now it works, the REST endpoint is open (not the CLI tho)
with a list of all names, each one can be queried individually like on https://talkshop.name

also should come with suggestions for setting up your own server with regard to CORS and self signed certificates (the option to remove SSL isn't working on this version)

### DIFF
--- a/app.go
+++ b/app.go
@@ -29,6 +29,7 @@ type nameserviceApp struct {
 
 	keyMain          *sdk.KVStoreKey
 	keyAccount       *sdk.KVStoreKey
+	keyNSallNames    *sdk.KVStoreKey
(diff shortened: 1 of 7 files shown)


okwme:
aaaaaand should probably come with an update to the actual tutorial

jackzampolin:
Lets hold of on merging this until we have the right way to do this. We also might want to address https://github.com/cosmos/sdk-application-tutorial/issues/1 as part of this work as well.

sunnya97:
Replaced by #44

PR #36, opened 2019-02-15
Delete --chain-id from nscli command in document.

yangyanqing:
Since the chain-id was saved by `nscli config chain-id testchain`, `--chain-id` was a redundancy parameter in command such as `nscli tx nameservice buy-name ...`.

```bash
# Configure your CLI to eliminate need for chain-id flag
nscli config chain-id testchain
```

### DIFF
--- a/tutorial/build-run.md
+++ b/tutorial/build-run.md
@@ -53,33 +53,25 @@ Open another terminal to run commands against the network you have just created:
 
 ```bash
 # First check the accounts to ensure they have funds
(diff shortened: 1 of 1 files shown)


jackzampolin:
Thank you for catching that @yangyanqing. Also thank you for the contribution!

yangyanqing:
You are welcome, i'll put more effort into cosmos from right now. :) @jackzampolin

PR #37, opened 2019-02-21
Small fixes (typos, comments) found running through the tutorial

t-bast:
Great job on the tutorial!
Here are the small typos/markdown stuff I ran against while following it (my markdown linter did most of the work).

I took some notes on what went through my mind / potential improvements, take them with a grain of salt but I hope it can be useful to make it even better:

**General**

The `Edit on Github` links don't seem to work (probably because you expect to move the tutorial inside the cosmos-sdk repo).

**Start your application**

"Commit if the ABCI message is DeliverTx" -> isn't this a bit misleading? DeliverTx should only be adding to the tx pool right? If DeliverTx is rejected or found invalid at block commit time bad things might happen.

Maybe mention that the user should do a `go get ./...` to fetch dependencies (or `go get github.com/tendermint/tendermint` and `go get github.com/cosmos/cosmos-sdk`). Later you introduce `dep`, but maybe it's worth introducing it at the start?

Codec not defined at that point so project not building, maybe it makes sense to start with the x/nameservice module and then create the app.go (because `MakeCodec()` isn't defined).

**BuyName**

`Bid.IsPositive()` doesn't exist anymore. I replaced by `Bid.IsAllPositive()`

**Codec File**

Maybe explain why the `WhoIs` struct doesn't need to be registered (because it's not included in any transaction and doesn't need canonical marshalling, it's just used in CLI RPC).

**Building and running the application**

This part has some redundant parts with the previous ones (about `dep`, building and installing).
I think they can be removed.

**Conclusion**

Maybe a small conclusion paragraph that lists some potentially interesting apps that can be built?
Or a list of known projects building on top of the Cosmos SDK?

### DIFF
--- a/tutorial/app-init.md
+++ b/tutorial/app-init.md
@@ -4,7 +4,12 @@ Get started by creating a new file: `./app.go`. This file is the heart of your d
 
 In `app.go`, you define what the application does when it receives a transaction. But first, it needs to be able to receive transactions in the correct order. This is the role of the [Tendermint consensus engine](https://github.com/tendermint/tendermint).
 
-Start by importing the necessary dependencies:
+Start by installing and importing the necessary dependencies:
+
+```sh
+go get -u github.com/tendermint/tendermint
+go get -u github.com/cosmos/cosmos-sdk
+```
 
 ```go
 package app
@@ -48,11 +53,12 @@ A couple of the packages here are `tendermint` packages. Tendermint passes trans
 Fortunately, you do not have to implement the ABCI interface. The Cosmos SDK provides a boilerplate implementation of it in the form of [`baseapp`](https://godoc.org/github.com/cosmos/cosmos-sdk/baseapp).
 
 Here is what `baseapp` does:
+
 - Decode transactions received from the Tendermint consensus engine.
 - Extract messages from transactions and do basic sanity checks.
 - Route the message to the appropriate module so that it can be processed. Note that `baseapp` has no knowledge of the specific modules you want to use. It is your job to declare such modules in `app.go`, as you will see later in this tutorial. `baseapp` only implements the core routing logic that can be applied to any module.
 - Commit if the ABCI message is [`DeliverTx`](https://tendermint.com/docs/spec/abci/abci.html#delivertx) ([`CheckTx`](https://tendermint.com/docs/spec/abci/abci.html#checktx) changes are not persistent).
-- Help set up [`Beginblock`](https://tendermint.com/docs/spec/abci/abci.html#beginblock) and [`Endblock`](https://tendermint.com/docs/spec/abci/abci.html#endblock), two messages that enable you to define logic executed at the beginning and end of each block. In practice, each module implements its own `BeginBlock` and `EndBlock` sub-logic, and the role of the app is to aggregate everything together (*Note: you won't be using these messages in your application*).
+- Help set up [`Beginblock`](https://tendermint.com/docs/spec/abci/abci.html#beginblock) and [`Endblock`](https://tendermint.com/docs/spec/abci/abci.html#endblock), two messages that enable you to define logic executed at the beginning and end of each block. In practice, each module implements its own `BeginBlock` and `EndBlock` sub-logic, and the role of the app is to aggregate everything together (_Note: you won't be using these messages in your application_).
 - Help initialise your state.
 - Help set up queries.
 
--- a/tutorial/app-init.md
+++ b/tutorial/app-init.md
@@ -4,12 +4,7 @@ Get started by creating a new file: `./app.go`. This file is the heart of your d
 
 In `app.go`, you define what the application does when it receives a transaction. But first, it needs to be able to receive transactions in the correct order. This is the role of the [Tendermint consensus engine](https://github.com/tendermint/tendermint).
 
-Start by installing and importing the necessary dependencies:
-
-```sh
-go get -u github.com/tendermint/tendermint
-go get -u github.com/cosmos/cosmos-sdk
-```
+Start by importing the necessary dependencies:
 
 ```go
 package app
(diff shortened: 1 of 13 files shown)


tutorial/app-init.md
```diff
@@ -4,7 +4,12 @@ Get started by creating a new file: `./app.go`. This file is the heart of your d
 
 In `app.go`, you define what the application does when it receives a transaction. But first, it needs to be able to receive transactions in the correct order. This is the role of the [Tendermint consensus engine](https://github.com/tendermint/tendermint).
 
-Start by importing the necessary dependencies:
+Start by installing and importing the necessary dependencies:
+
+```sh
```
yangyanqing: I don't think the following two steps were necessary, because they would be call while running `make get_vendor_deps`.

t-bast: Once we reach the end of the tutorial (where we start using deps) I agree :)
But it was a bit weird to follow the tutorial with something that couldn't build, which is why I added that...
Maybe the dep installation should be in the first steps of the tutorial (but that should probably be done in another PR)?

jackzampolin: I don't think we need this here.

t-bast: Allright I'll remove it then, maybe another change to the tutorial can bring `dep` closer to the introduction.

t-bast:
Sure I'll re-configure my editor and remove the whitespace changes, it makes more sense to be in a dedicated PR of its own.

tutorial/cli.md
```diff
@@ -211,37 +212,38 @@ func NewModuleClient(storeKey string, cdc *amino.Codec) ModuleClient {
 
 // GetQueryCmd returns the cli query commands for this module
 func (mc ModuleClient) GetQueryCmd() *cobra.Command {
-	// Group gov queries under a subcommand
-	govQueryCmd := &cobra.Command{
+	// Group nameservice queries under a subcommand
```
jackzampolin: Thank you for catching my copy pasta here 🙈

jackzampolin: If you are making changes in the code in the `.md` files, please also make the same change in the corresponding `.go` file!

tutorial/buy-name.md
```diff
@@ -21,27 +21,27 @@ func NewMsgBuyName(name string, bid sdk.Coins, buyer sdk.AccAddress) MsgBuyName
 	}
 }
 
-// Type Implements Msg.
+// Route should return the name of the module
```
jackzampolin: If you are making a change in the code in the `.md` files, please make a corresponding change in the `.go` file.

tutorial/entrypoint.md
```diff
@@ -250,10 +250,8 @@ func SimpleAppGenTx(cdc *codec.Codec, pk crypto.PubKey) (
 ```
 
 Notes on the above code:
-- Most of the code above combines the CLI commands from
-	1. Tendermint
-	2. Cosmos-SDK
-	3. Your Nameservice module
+
+- Most of the code above combines the CLI commands from 1. Tendermint 2. Cosmos-SDK 3. Your Nameservice module
```
jackzampolin: You can remove the numbers here.

tutorial/entrypoint.md
```diff
@@ -425,10 +423,8 @@ func initConfig(cmd *cobra.Command) error {
 ```
 
 Note:
-- The code combines the CLI commands from:
-	1. Tendermint
-	2. Cosmos-SDK
-	3. Your Nameservice module
+
+- The code combines the CLI commands from: 1. Tendermint 2. Cosmos-SDK 3. Your Nameservice module
```
jackzampolin: ditto

tutorial/queriers.md
```diff
@@ -17,7 +17,7 @@ import (
 	abci "github.com/tendermint/tendermint/abci/types"
 )
 
-// query endpoints supported by the governance Querier
+// query endpoints supported by the nameservice Querier
```
jackzampolin: If you are making a change in the code in the `.md` files, please make a corresponding change in the `.go` file.

tutorial/set-name.md
```diff
@@ -60,10 +61,10 @@ func (msg MsgSetName) ValidateBasic() sdk.Error {
 }
 ```
 
-`ValidateBasic` is used to provide some basic **stateless** checks on the validity of the `Msg`.  In this case, check that none of the attributes are empty. Note the use of the `sdk.Error` types here. The SDK provides a set of error types that are frequently encountered by application developers.
+`ValidateBasic` is used to provide some basic **stateless** checks on the validity of the `Msg`. In this case, check that none of the attributes are empty. Note the use of the `sdk.Error` types here. The SDK provides a set of error types that are frequently encountered by application developers.
 
 ```go
-// GetSignBytes Implements Msg.
+// GetSignBytes encodes the message for signing
```
jackzampolin: If you are making a change in the code in the `.md` files, please make a corresponding change in the `.go` file.

tutorial/set-name.md
```diff
@@ -115,12 +116,12 @@ func NewHandler(keeper Keeper) sdk.Handler {
...
 `NewHandler` is essentially a sub-router that directs messages coming into this module to the proper handler. At the moment, there is only one `Msg`/`Handler`.
 
-Now, you need to define the actual logic for handling the `MsgSetName` message in `handleMsgSetName`.:
+Now, you need to define the actual logic for handling the `MsgSetName` message in `handleMsgSetName`:
 
 > _*NOTE*_: The naming convention for handler names in the SDK is `handleMsg{ .Action }`
 
 ```go
-// Handle MsgSetName
+// Handle a message to set name
```
jackzampolin: If you are making a change in the code in the `.md` files, please make a corresponding change in the `.go` file.

tutorial/buy-name.md
t-bast: That does sound logical ;)

jackzampolin:
Thank you for fixing that @t-bast!

PR #40, opened 2019-03-03
Update app-complete.md

EG-easy:
just fixed tutorial code typo to import package


jackzampolin:
@EG-easy Thank you for the fix!

PR #41, opened 2019-03-06
Combined three stores into one with Whois struct as value

### LINKED ISSUE #1: Proposal: Collapsing the 3 stores
Currently there are three stores for the nameservice module. I think we should collapse them into one. This would have the following benefits:

- Mirror current best practices in existing modules
- Add a `types.go` file to the module for the struct that would contain `{value, price, owner}`, which would help readers understand its use. They will likely have to have one if they ever build their own module. 
- Unmarshalling overhead is ok since we always use at least 2/3 fields of the struct in functions.

sunnya97:
closes #1

### DIFF
--- a/x/nameservice/types.go
+++ b/x/nameservice/types.go
@@ -1,0 +1,12 @@
+package nameservice
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// Whois is a struct that contains all the metadata of a name
+type Whois struct {
+	Value string         `json:"value"`
+	Owner sdk.AccAddress `json:"owner"`
+	Price sdk.Coins      `json:"price"`
+}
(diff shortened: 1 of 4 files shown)


x/nameservice/types.go
```diff
@@ -0,0 +1,22 @@
+package nameservice
```
jackzampolin: Like this whole file.

jackzampolin:
This looks great @sunnya97 we need to update the text of the tutorial for this change and that needs to be in this PR.

okwme:
updated the markdown files on sunny's branch. If he merges those this PR might be ok to merge too?
https://github.com/sunnya97/sdk-application-tutorial/pull/1

jackzampolin:
Lets go ahead and merge when ready @okwme

sunnya97:
Replaced by #43

PR #42, opened 2019-03-06
All names iterator

sunnya97:
replaces #29 

Extends #41

### DIFF
--- a/x/nameservice/querier.go
+++ b/x/nameservice/querier.go
@@ -40,22 +40,11 @@ func queryResolve(ctx sdk.Context, path []string, req abci.RequestQuery, keeper 
 	return []byte(value), nil
 }
 
-// Whois represents a name -> value lookup
-type Whois struct {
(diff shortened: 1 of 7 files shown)


x/nameservice/querier.go
```diff
@@ -64,3 +56,22 @@ func queryWhois(ctx sdk.Context, path []string, req abci.RequestQuery, keeper Ke
 
 	return bz, nil
 }
+
+// nolint: unparam
+func queryNames(ctx sdk.Context, req abci.RequestQuery, keeper Keeper) (res []byte, err sdk.Error) {
```
jackzampolin: We may want to think about pagination here cc @okwme

jackzampolin:
This looks great so far. Be sure to also update the `.md` files that correspond to the `.go` files @sunnya97

sunnya97:
Replaced by #44

PR #45, opened 2019-03-07
R4R: Json and Stringer structs in querier instead of CLI

sunnya97:


jackzampolin:
This is a great add @sunnya97!

PR #58, opened 2019-04-02
Three stores had been combined in #43.

yangyanqing:
So we should delete `nameStore` and ` ownerStore`.


jackzampolin:
I think we need to rewrite an entire section here to better explain what is happening. Also the fact that we should explain why `WhoIs` doesn't need to be `amino` encoded.

jackzampolin:
This also needs a rebase

yangyanqing:
> This also needs a rebase

Done!

> I think we need to rewrite an entire section here to better explain what is happening.

Did you mean why we combine three store to one ?

> Also the fact that we should explain why `WhoIs` doesn't need to be `amino` encoded.

I want to know too.

PR #63, opened 2019-04-05
R4R: Translate document of ledger to chinese. Related: #62

yangyanqing:
related: #62


yangyanqing:
Is there any change I should do ? @jackzampolin

PR #69, opened 2019-04-12
CMD run fixed

NicolasLagaillardie:
On linux, you need to run exec with ./yourapp

### DIFF
--- a/tutorial/en/build-run.md
+++ b/tutorial/en/build-run.md
@@ -9,8 +9,8 @@ If you want to build the `nameservice` application in this repo to see the funct
 make install
 
 # Now you should be able to run the following commands:
-nsd help
-nscli help
+./nsd help
+./nscli help
 ```
 
 ## Running the live network and using the commands


tutorial/en/build-run.md
```diff
@@ -9,8 +9,8 @@ If you want to build the `nameservice` application in this repo to see the funct
 make install
 
 # Now you should be able to run the following commands:
-nsd help
-nscli help
+./nsd help
```
yangyanqing: `nsd` and `nscli` would be create in folder`${GOBIN}`. 
After the following steps, `${GOBIN}` was already appended to ${PATH}. 
So, we can run `nscli` directly, `./` is redundancy

```bash
mkdir -p $HOME/go/bin
echo "export GOPATH=$HOME/go" >> ~/.bash_profile
echo "export GOBIN=\$GOPATH/bin" >> ~/.bash_profile
echo "export PATH=\$PATH:\$GOBIN" >> ~/.bash_profile
echo "export GO111MODULE=on" >> ~/.bash_profile
source ~/.bash_profile
```

PR #70, opened 2019-04-12
Var was defined twice

NicolasLagaillardie:
"github.com/cosmos/cosmos-sdk/x/bank" was defined twice, creating issues


alexanderbez:
There is a merge conflict here. Also, ideally changes should be targeted against `develop`.

marbar3778:
This has been fixed, since the PR was opened, also since the most recent refactor the docs were updated, Thank you for the contribution

PR #78, opened 2019-04-27
md fix: add a space to match md syntax

hello2mao:
**What this PR does / why we need it:**
Add a space to match md syntax for a better view.


yangyanqing:
I had searched all docs with `^#+[^ #]+`, only one line without space between `#` and title.

PR #81, opened 2019-05-03
use application codec not json encoder to fix type / value vulnerability

mjackson001:


jackzampolin:
Thanks for this PR @mjackson001! Because there were merge conflicts here I fixed this in https://github.com/cosmos/sdk-application-tutorial/pull/97

PR #89, opened 2019-05-24
Add instructions for Talkshop Name Server

onggunhao:
Not sure if you guys want it to be part of the tutorial, but these additional pointers were useful for me setting up Talkshop name server. Can do the CN translation later on if it helps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,5 +10,12 @@ This repository contains the source code of the nameservice tutorial.
 
 **[Click here](./tutorial/build-run.md)**  for instructions on how to build and run the code.
 
+## Talkshop Name Server
+
+There is a public [Talkshop Name Server](https://talkshop.name/) where you can interact with a live instance of namechain. Instructions (and faucet) are available at [https://talkshop.name/](https://talkshop.name/)
+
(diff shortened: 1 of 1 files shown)


jackzampolin:
I don't know if we want to commit to keeping that server up. Best not to mention it IMO. Thoughts @okwme ?

onggunhao:
For what it's worth, it was pretty cool to connect to the Talkshop name server rather than just my local testnet. Hope you'll keep it around (at least for the early days of developer adoption!)

okwme:
I'd hate to have to start monitoring it for inappropriate content if it got too much broad attention but agree the web server is a really fun finale to the tutorial. Maybe for now it's better to introduce the server only to in-person workshop participants and try to find a long term solution that doesn't need monitoring?

I'd also like to write a front-end tutorial for building something like the talkshop.name web app. Maybe allowing a user to spin up the front-end themselves is an eventual solution?

onggunhao:
@okwme A frontend tutorial would be really interesting, hope you work on it

marbar3778:
will add the current FE as a docker container in the sdk-tutorial, so people can run the tutorial with the forntend. Will close this PR in favor of that.

PR #91, opened 2019-05-30
app-init.md fixes

theodesp:
I'm following the tutorial and I noticed that the app-init.md part was missing some imports. Also the name `Cosmos` is spelled wrong. I added the missing imports so anyone that follows the tutorial will not have issues.


marbar3778:
This has been fixed in the most recent merge with the updated version of the tutorial

PR #92, opened 2019-06-03
Implement new DeleteName Message Type

aayushijain23:
Added a new message called DeleteName to this tutorial. The following changes have been made:
1) Added a new message type for DeleteName
2) Updated the keeper with a new method for deleting the Whois struct for a particular name.
3) Updated the Handler with what actions to be taken for a DeleteName message.
4) Registered the DeleteName interface in Codec.
5) Added a cli command for the DeleteName transaction and added corresponding changes in the module client.
6) Finally added a REST endpoint for the same.


x/nameservice/handler.go
```diff
@@ -50,3 +52,12 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
 	keeper.SetPrice(ctx, msg.Name, msg.Bid)
 	return sdk.Result{}
 }
+
+// Handle a message to delete name
+func handleMsgDeleteName(ctx sdk.Context, keeper Keeper, msg MsgDeleteName) sdk.Result {
+	if !msg.Owner.Equals(keeper.GetOwner(ctx, msg.Name)) { // Checks if the the msg sender is the same as the current owner
+		return sdk.ErrUnauthorized("Incorrect Owner").Result() // If not, throw an error
```
alexanderbez:
```suggestion
		return sdk.ErrUnauthorized("Incorrect Owner").Result()
```

x/nameservice/handler.go
```diff
@@ -50,3 +52,12 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
 	keeper.SetPrice(ctx, msg.Name, msg.Bid)
 	return sdk.Result{}
 }
+
+// Handle a message to delete name
+func handleMsgDeleteName(ctx sdk.Context, keeper Keeper, msg MsgDeleteName) sdk.Result {
+	if !msg.Owner.Equals(keeper.GetOwner(ctx, msg.Name)) { // Checks if the the msg sender is the same as the current owner
```
alexanderbez:
```suggestion
       // Checks if the the msg sender is the same as the current owner
	if !msg.Owner.Equals(keeper.GetOwner(ctx, msg.Name)) {
```

x/nameservice/handler.go
```diff
@@ -50,3 +52,12 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
...
+		return sdk.ErrUnauthorized("Incorrect Owner").Result() // If not, throw an error
+	}
+	keeper.DeleteName(ctx, msg.Name) // If so, delete the name specified in the msg.
```
alexanderbez:
```suggestion

	keeper.DeleteName(ctx, msg.Name)
```

x/nameservice/handler.go
```diff
@@ -50,3 +52,12 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
...
+	return sdk.Result{}              // return
```
alexanderbez:
```suggestion
	return sdk.Result{}
```

x/nameservice/keeper.go
```diff
@@ -46,6 +46,15 @@ func (k Keeper) SetWhois(ctx sdk.Context, name string, whois Whois) {
 	store.Set([]byte(name), k.cdc.MustMarshalBinaryBare(whois))
 }
 
+// Deletes the entire Whois metadata struct for a name
+func (k Keeper) DeleteWhois(ctx sdk.Context, name string) {
```
alexanderbez: How is this different that `DeleteName`? Why not just have a single method `DeleteWhois` or `DeleteName`. We should also return an error if the name doesn't exist.

aayushijain23: You're right. There isn't any difference between the two methods. Actually, Getter and Setter methods were defined in the form of SetWhoIs and GetWhoIs. And then additional methods were added for getting specific parameters from the store based on the name, such as: SetOwner(), GetOwner() etc. I thought of maintaining consistency by following the same structure. But yeah in the case of deleting the name, there aren't any additional methods that can be specified. Will keep `DeleteWhoIs`. And yeah I'll return an error if the name doesn't exist.

tutorial/cli.md
```diff
@@ -217,6 +218,34 @@ func GetCmdSetName(cdc *codec.Codec) *cobra.Command {
...
+				return err
+			}
+
+			msg := types.NewMsgDeleteName(args[0], cliCtx.GetFromAddress())
+			err := msg.ValidateBasic()
+			if err != nil {
+				return err
+			}
+
+			cliCtx.PrintResponse = true
```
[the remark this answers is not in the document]
aayushijain23: I thought this prints the response in the command line, when we test the code via cli commands. Will remove it.

tutorial/delete-name.md
```diff
@@ -73,14 +73,20 @@ Finally, define the `DeleteName` `handler` function which performs the state tra
 ```go
 // Handle a message to delete name
 func handleMsgDeleteName(ctx sdk.Context, keeper Keeper, msg MsgDeleteName) sdk.Result {
-	if !msg.Owner.Equals(keeper.GetOwner(ctx, msg.Name)) { // Checks if the the msg sender is the same as the current owner
-		return sdk.ErrUnauthorized("Incorrect Owner").Result() // If not, throw an error
+	// Checks if the name is present in the store or not
```
alexanderbez: This comment is redundant imho

tutorial/delete-name.md
```diff
@@ -73,14 +73,20 @@ Finally, define the `DeleteName` `handler` function which performs the state tra
...
+	if !keeper.IsNamePresent(ctx, msg.Name) {
+		return types.ErrNameDoesNotExist(types.DefaultCodespace).Result()
 	}
-	keeper.DeleteWhois(ctx, msg.Name) // If so, delete the name specified in the msg.
-	return sdk.Result{}              // return
+	// Checks if the the msg sender is the same as the current owner
```
alexanderbez: ditto

tutorial/delete-name.md
```diff
@@ -73,14 +73,20 @@ Finally, define the `DeleteName` `handler` function which performs the state tra
...
+	if !msg.Owner.Equals(keeper.GetOwner(ctx, msg.Name)) { 
+		return sdk.ErrUnauthorized("Incorrect Owner").Result() 
+	}
+	// If so, delete the name specified in the msg.
```
alexanderbez: ditto

x/nameservice/handler.go
```diff
@@ -55,7 +56,11 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
 
 // Handle a message to delete name
 func handleMsgDeleteName(ctx sdk.Context, keeper Keeper, msg MsgDeleteName) sdk.Result {
-    // Checks if the the msg sender is the same as the current owner
+	// Checks if the name is present in the store or not
```
alexanderbez: ditto

x/nameservice/handler.go
```diff
@@ -55,7 +56,11 @@ func handleMsgBuyName(ctx sdk.Context, keeper Keeper, msg MsgBuyName) sdk.Result
...
+	if !keeper.IsNamePresent(ctx, msg.Name) {
+		return types.ErrNameDoesNotExist(types.DefaultCodespace).Result()
+	}
+    // Checks if the msg sender is the same as the current owner
```
alexanderbez: ditto

aayushijain23: Cool, will remove the unnecessary comments.

tutorial/cli.md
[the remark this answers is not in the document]
alexanderbez: This still needs to be removed.

tutorial/delete-name.md
```diff
@@ -0,0 +1,89 @@
+# Delete Name
+
+## Msg
+
+Now it is time to define the `Msg` for deleting names and add it to the `./x/nameservice/types/msgs.go` file. This code is very similar to `SetName`:
+
+```go
+// MsgDeleteName defines a DeleteName message
+type MsgDeleteName struct {
+	Name  string		 `json:"name"`
```
alexanderbez: Formatting is off.

aayushijain23: Didn't quite understand this one. I followed the same format as buy-name.md and set-name.md. Everywhere I have noticed that the `json` names are structured in a column, one after the other. Hence the code looks as follows:
```
type MsgDeleteName struct {
	Name  string         `json:"name"`
	Owner sdk.AccAddress `json:"owner"`
}
```
Since sdk.AccAddress as a term takes more space than string, the term `json:"name"` is pushed to align with `json:"owner"`

alexanderbez: The formatting of Go is not correct. See: https://github.com/cosmos/sdk-application-tutorial/blob/88709734312254089c3c1e3ff5935b436fce4cba/tutorial/delete-name.md

The field tags need to be aligned. You can just copy-paste this from the actual gocode.

aayushijain23: Oh I thought you meant the alignment of the field tags in the code is not right. Yeah I couldn't see the preview before pushing the code. I'll copy it from there.

aayushijain23: On a side note, found the preview option in vscode.

PR #100, opened 2019-06-18
Fix broken link in Cosmos sdk tutorial

michaeleliot:
Looks like the develop branch isn't there anymore. Might be better to have the link point to master instead!


marbar3778:
Thank you!

PR #111, opened 2019-06-29
fix cmd/nsd/main.go

amazingandyyy:
apparently https://github.com/cosmos/cosmos-sdk/tree/master/x/genaccounts is in "github.com/cosmos/cosmos-sdk/x/genaccounts"

related to https://github.com/cosmos/sdk-application-tutorial/pull/110

### DIFF
--- a/cmd/nsd/main.go
+++ b/cmd/nsd/main.go
@@ -5,8 +5,8 @@ import (
 	"io"
 
 	"github.com/cosmos/cosmos-sdk/server"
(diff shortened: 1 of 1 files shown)


marbar3778:
with the version of the master branch that is being used by the tutorial, this does not work, you would have to update to a more recent version of master.

marbar3778:
updating the tutorial in this PR: https://github.com/cosmos/sdk-application-tutorial/pull/120, closing this in favor of it as it has the most recent version of the Cosmos-SDK

PR #117, opened 2019-07-13
Original rest.go file was updated but rest.md has not been updated

HyoungsungKim:


marbar3778:
Thank you for your contribution

PR #133, opened 2019-08-21
update to 0.37

marbar3778:
- [x] make changes to code
- [x] update docs
- [x] change file structure in tutorial readme


marbar3778:
I do agree with you @hschoenburg, but I was following https://github.com/cosmos/cosmos-sdk/blob/master/docs/modules/SPEC.md. If this isnt something we want to enforce or only enforce in some places then i can change it back to similar as before

hschoenburg:
> I do agree with you @hschoenburg, but I was following https://github.com/cosmos/cosmos-sdk/blob/master/docs/modules/SPEC.md. If this isnt something we want to enforce or only enforce in some places then i can change it back to similar as before

Ah ok I didnt see that in the spec. It probably does help.

hschoenburg:
oops didnt mean to close it just there. Approving now.

marbar3778:
thank you, @hschoenburg

PR #136, opened 2019-09-07
Add message tests

marbar3778:
- step 1 of adding testing to the tutorial

ref #13 

Signed-off-by: Marko Baricevic <marbar3778@yahoo.com>

### DIFF
--- a/x/nameservice/internal/types/msgs_test.go
+++ b/x/nameservice/internal/types/msgs_test.go
@@ -7,20 +7,10 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-var (
-	name   = "maTurtle"
-	value  = "1"
-	acc    = sdk.AccAddress([]byte("me"))
-	name2  = "a"
-	value2 = "2"
-	acc2   = sdk.AccAddress([]byte("you"))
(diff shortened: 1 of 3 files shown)


x/nameservice/internal/types/msgs_test.go
```diff
@@ -0,0 +1,148 @@
...
+	"testing"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/stretchr/testify/require"
+)
+
+func TestMsgSetName(t *testing.T) {
+	name := "maTurtle"
+	value := "1"
+	acc := sdk.AccAddress([]byte("me"))
```
okwme: not clear to me what's going on here. If you instantiate an address using an arbitrary byte string is that used as the private key?

marbar3778: this is just creating a dummy account for testing. there is no need fo the priv key here, once we do simulaiton tests we will need to integrate something with priv key i believe

hschoenburg: i think its ok not to use valid addresses here since the test is focused on the nameservice message

PR #182, opened 2019-11-01
Fix tutorial for cn

yjhmelody:
I find that the cn tutorial is very old. So I try to update the code and some minor descriptions.
Any way to quote code just like EN tutorial?


marbar3778:
Thank you for your contribution

PR #184, opened 2019-11-03
Code/compile fixes

quantaf1re:
Changed:
 - Spelling
 - Mentioning that nothing is supposed to be cloned
 - Saying to create files instead of open them when they don't exist
 - Added the comma so this can compile:
```
    var app = &helloChainApp{
        appStarter,
    }
```
 - Moving the `make.md` file to near the end because it requires that the cli is built
 - Corrected file path to `x/greeter/client/cli/query.go` in `client.md`


okwme:
Love the PR thank you!!!

I'm curious to hear @hschoenburg 's thoughts about the re-ordering of the tutorial though. I think he had that order in mind to show what a completely basic chain could do _before_ showing what extra functionality could add. 

Might make sense to switch it up, but if so we should keep an eye out for continuity errors like here:
https://github.com/cosmos/sdk-tutorials/commit/a4cd78c267040b1f7f71496fce05370f1b5c30a5#diff-336e16a561818fb034c8173b054e0deaL106-R107
```
... Now lets add some functionality of our own making to see how flexible
the Cosmos SDK can be. 
```

quantaf1re:
Ok, I added another PR without the restructuring

quantaf1re:
Sorry, I meant that I removed the restructuring from this PR and added another PR that includes the restructuring

hellochain/tutorial/03-make.md
```diff
@@ -24,7 +24,6 @@ all: install
 
 install: go.sum
 	go install $(BUILD_FLAGS) ./cmd/hcd
-	go install $(BUILD_FLAGS) ./cmd/hccli
```
marbar3778: I do believe this should stay here

quantaf1re: The order of the tutorial makes it such that `hccli` doesn't exist at the point the user `make install`s this make file - that's why I originally reordered the files, which is now in https://github.com/cosmos/sdk-tutorials/pull/185

PR #197, opened 2019-11-21
fix misspellings and inconsistent (stale?) english

alexkogon:


nameservice/tutorial/21-build-run.md
```diff
@@ -1,143 +1,143 @@
----
```
gamarin2: not sure why the diff is that big on this file?

alexkogon: Whoa didn't notice that; no idea it looks exactly the same why is it showing as all changed?

Issue #195, opened 2019-11-21
Mismatch in instructions and the code in the nameservice tutorial


hammadtq:
The instructions at [06-keeper](https://tutorials.cosmos.network/nameservice/tutorial/06-keeper.html) asks you to create a new file at `./x/nameservice/keeper.go` while in [12-alias](https://tutorials.cosmos.network/nameservice/tutorial/12-alias.html) the tutorial is trying to import `github.com/cosmos/sdk-tutorials/nameservice/x/nameservice/internal/keeper`.

This is a general structure and import issue in all of the tutorial as in the [nameservice source code](https://github.com/cosmos/sdk-tutorials/tree/master/nameservice/x/nameservice) we have an `internal` folder inside the module but in the tutorial we are not creating one but referencing to it.

marbar3778:
ah thank you, I apoligize for the confusion, we are finishing work a scaffolding tool to take of this for you. We hop to have it out tomorrow or monday

mcadam:
any updates? what are the paths we should use in the alias.go file?

mcadam:
There was quite a bit more errors than that commit, I had to fix other paths in other files too, in the tutorial the keeper and queryier are put in the package keeper, at different paths than internal, etc.

It would be nice for  someone to follow the tutorial fully https://tutorials.cosmos.network/nameservice/tutorial/00-intro.html
and then fix the errors at the end while trying to build and make sure everything works. 
In my opinion its quite important when approaching a new project like this and follow a tutorial to be able to go all the way, otherwise you lose potential users :)

PR #201, opened 2019-12-02
Further fixes to nameservice tutorial

marbar3778:
- further changes for #195

Signed-off-by: Marko Baricevic <marbar3778@yahoo.com>


nameservice/tutorial/00-intro.md
```diff
@@ -63,9 +62,10 @@ Through the course of this tutorial you will create the following files that mak
 Start by creating a new git repository:
 
 ```bash
-mkdir -p $GOPATH/src/github.com/{ .Username }/nameservice
-cd $GOPATH/src/github.com/{ .Username }/nameservice
+mkdir -p nameservice
```
okwme:
```suggestion
mkdir nameservice
```

would suggest removing p, without multiple ply in the tree it is only preventing errors when the dir already exists (an error you would presumably want here)

PR #224, opened 2020-01-17
Updated theme to the latest version

fadeev:


hellochain/tutorial/03-types.md
```diff
@@ -10,4 +10,5 @@ define this struct and other necessary components in a types file.
 
 Save this to `x/greeter/internal/types/types.go`
 
-<<< @/hellochain/x/greeter/internal/types/types.go
++++ https://github.com/cosmos/sdk-tutorials/blob/master/hellochain/x/greeter/internal/types/types.go
```
hschoenburg: hmm relative paths dont work anymore? we have to use URLs? why is that if you dont mind me asking

hellochain/tutorial/15-full-app.md
```diff
@@ -11,4 +11,4 @@ and add greeter's AppModuleBasic and AppModule alongside the other modules.
 
 Update your `app.go` to look like the following
 
-<<< @/hellochain/app.go{12,25,26,33,36,39,44,45,49,50,53}
```
hschoenburg: is there anyway we can keep the line highlights? they are pretty useful

hellochain/tutorial/03-types.md
fadeev: We didn't use relative paths in the SDK docs, so this feature got kind of forgotten. I've added it back and modified MDs accordingly.

hellochain/tutorial/15-full-app.md
fadeev: We're using `{...}` syntax for annotating things like synopsis and prerequisite links, so we can't use it for line numbers. I'm thinking of using `.../app.go[12,25,26]` syntax instead.

PR #230, opened 2020-01-21
wrong link

okwme:


okwme:
also removes link that shouldn't be there

Issue #221, opened 2020-01-17
Simplify Hellochain Tutorial


ghost:
I landed on a tutorial:

https://tutorials.cosmos.network/hellochain/tutorial/00-intro.html

It starts a bitoverwhelming, like this:

```sh
mkdir -p $GOPATH/src/github.com/{ .Username }/hellochain
cd $GOPATH/src/github.com/{ .Username }/hellochain
git init
```

which is not very convenient (possibly google's go fault, but still).

## Expected

Simplify this in the way:

### Prerequisites

* Go
* git

### Setup

Within your usual project location:

```sh
mkdir hellochain
cd hellochain
git init
```

marbar3778:
we need to rewrite the tutorial to use our scaffolding tool. I or my colleagues will spend time on this in the coming days

lazaridiscom:
> we need to rewrite the tutorial to use our scaffolding tool. 

The Idea of "doing everything by hand" is a good one. Personally, I don't like scaffolding very much, as it avoids the need to do decent reusable design. Though "scaffolding" does not exclude "good framework design", if done right.

Note that as a newcomer, I have the benefit of seeing barriers which the existent team does not see anymore. I'll provide a little bit more feedback within the next 48h.


## Possible Spawning Issue

* Allow for easy "project" creation within any location on filesystem.

lazaridiscom:
> we need to rewrite the tutorial to use our scaffolding tool. I or my colleagues will spend time on this in the coming days

I reached https://tutorials.cosmos.network/hellochain/tutorial/12-make.html, i am on Windows 10, and need here to abort (too lazy to work out the "missing make" barrier). Possibly I retry on an virtual machine later.

Scaffolding should do it, i guess, but a "prerequisite" section should be added, too.

As an example for "straight to the point", see this EOS starter:

https://github.com/lazaridiscom/dapp-platform/tree/master/eos

You must be clear about your target-audience (e.g.: professional devs & evaluators, casual tech-kids which want to play with crypto => all is fine, just state it. I tend to say that addressing the pro's should be prioritized)

lazaridiscom:
@marbar3778 , re "we need to rewrite the tutorial to use our scaffolding tool", it seems not necessary, I noticed that code is in the repo, so if one wishes to do it faster => clone repo, go through tuturial and be happy that files are there.

Will most possibly try this tomorrow (on Windows WSL, see  #228), and comment.

lazaridiscom:
[...]

PR #238, opened 2020-01-24
provide a setup document, re #221

lazaridiscom:
Adding basic setup instructions.

Going through `hellochain` becomes now as easy as in:

```sh
git clone https://github.com/cosmos/sdk-tutorials.git
cd sdk-tutorials/hellochain
make install
hcd
hccli
```

then continue with the "try-it-out" section.

=> zero hassle

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 _[The code for these tutorials (and this site) can be found here on Github](https://www.github.com/cosmos/sdk-tutorials)_
 
+## Prerequisites
+
+After a quick [Development Environment Setup](SETUP.md), you're ready to select a tutorial from the list below.
+
 ## [Hellochain](./hellochain/tutorial/00-intro.md)
 
 This is the quick and simple "Hello World" introduction to building with the Cosmos SDK. You will build a fully-funcitonal blockchain with payment functionality and a custom "greeter" module. This tutorial makes use of the `utils/starter` package to reduce boilerplate and allow you to focus on the core concepts. Start here if you are evaluating Cosmos as your possible platform of choice.
--- a/SETUP.md
+++ b/SETUP.md
@@ -1,0 +1,27 @@
+# Development Environment Setup
+
+## Supported OS
+
+The example instructions are for Ubuntu 18.04 and for any other OS using ubuntu on a virtual machine (Windows: WSL, VirtualBox, Docker.
+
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -4,6 +4,10 @@
 
 _[The code for these tutorials (and this site) can be found here on Github](https://www.github.com/cosmos/sdk-tutorials)_
 
+## Prerequisites
+
+After a quick [Development Environment Setup](SETUP.md), you're ready to select a tutorial from the list below.
```
marbar3778: I think it may be better to link to the official go download page instead of providing a doc that says how to install it.

lazaridiscom: No way. Setup.md contains extracted explanations (the overall task took me hours), some specific to the tutorials (e.g. windows status).

Linking simply to a download page is sloppy.

marbar3778: we have had this discussion multiple times. The decision was to link to the go install page. this is how it is done in Tendermint, Gaia and the Cosmos-sdk.

lazaridiscom: Well, if you're all proud of your result, then simply shred this PR.

marbar3778: I definitely there is a way we can work together to get this pull request merged

SETUP.md
```diff
@@ -0,0 +1,27 @@
+# Development Environment Setup
+
+## Supported OS
+
+The example instructions are for Ubuntu 18.04 and for any other OS using ubuntu on a virtual machine (Windows: WSL, VirtualBox, Docker.
```
marbar3778: What if we rewrite some of this to be focused toward windows but link to go install page for mac and Linux?

lazaridiscom: Just merge it for now. If you (or someone else) can come up later with cli steps for mac, just add them. Similar with windows, but that seems to be high effort.

I added the link to the official download, so everyone should be happy.

(note that at this point, I'll most possibly not invest more time into this, you can take on if you want)

README.md
lazaridiscom: see https://github.com/cosmos/sdk-tutorials/pull/238#discussion_r370647498

SETUP.md
lazaridiscom: (though maybe I give it tomorrow one more pass, I just had an idea to simplify further)

PR #241, opened 2020-01-27
update: tutorials to 0.38 sdk

### LINKED ISSUE #72: Add nsdebug to cmd and makefile
Just like `gaiadebug`. It would be nice so that people can query txs bytes from tendermint rpc and then unmarshall them to JSON.

### LINKED ISSUE #131: Querying transaction hash via REST server
Am I correct to believe that [rest server](https://github.com/cosmos/sdk-application-tutorial/blob/master/tutorial/run-rest.md) that is started using `nscli` is meant to roughly mimic  [gaicli rest server](https://cosmos.network/docs/clients/service-providers.html#setting-up-the-rest-server)? In particular, should it fully support at least ICS0 Tendermint APIs on this [Swagger page](https://cosmos.network/rpc/)?

The API that is missing is `/txs/{hash}` (Get a Tx by hash). I am able to use this API via `gaicli` rest server but not via `nscli` rest server.

### LINKED ISSUE #140: Things that need to be updated
- [x] `genesis.go` should be split in two -> A `genesis.go` should be created in `/types`, it would contain the type definition of `GenesisState` as well as `DefaultGenesis` and `ValidateGenesis`
- [x] Add `expected_keepers.go` in `/types`, modify `keeper.go` accordingly.

marbar3778:
- [x] nameservice
- [x] hellochain
- [x] scavange

- closes #131 
- closes #72 
- closes #140


okwme:
HUGE update (THANK YOU) only natural that you forgot to _add the scavenge module_ (lol)
Is it possible to run an instance of each chain in CI to see if it starts?
I removed the daemon from all the `init.sh` scripts because it will overwrite the important mnemonic info, but maybe could add it back if it was just meant for confirming the chain runs?

marbar3778:
yea we can do something like that. It would be useful for sure. I will look into it.

marbar3778:
https://github.com/cosmos/sdk-tutorials/issues/243

PR #244, opened 2020-02-03
Update namespace tutorial to use errors package.

jgimeno:


jgimeno:
Ah wait, there is something wrong then. I was following this tutorial and somehow it has the old types. Let me check.

jgimeno:
What is the correct one actually? The sdkerrors package?

jgimeno:
I was using sdk version v0.37.4, I see that with 0.38 it works! I was super confused! There was a major update that caught me doing the Nameservice tutorial. We can close this one I think.

marbar3778:
ah my bad. If you have any questions please let me know. Ill do my best to answer

jgimeno:
There is still some unconsistency with the errors in the tutorial though, I am trying to understand it.

Example

https://github.com/cosmos/sdk-tutorials/blame/master/nameservice/tutorial/09-set-name.md#L127

This in the last version of the sdk does not exist. But if I use the sdkerrors I cannot use .Result() to return an sdk.Result

jgimeno:
Sorry, my fault, now I see that the "correct" is the one included in the x/ module of this same repository, let me update it on the tutorial. 

Let's keep this PR opened so I can update it given that I am following the tutorial.

jgimeno:
Now I think is correct @marbar3778, sorry for all the commits and reverts.

marbar3778:
na this is awesome, thank you!! this reminds me i finally figured out how to show code through markdown without copy paste, we shouldnt have this problem in the future

PR #251, opened 2020-02-12
deps: update sdk

marbar3778:


marbar3778:
yea this has the fix of the pruning stuff so it was easier

PR #279, opened 2020-03-26
fix inconsistencies with nameservice

marbar3778:
closes #275 

fix some inconsistencies with nameservice and tutorial.

some of the changes are being made in scaffold to complete the issue.

PR #280, opened 2020-03-27
add happy path to nameservice

okwme:
Added a happypath bash command that can be used in lieu of actual tests. Just meant for a sanity check that the "happy path" use of the tutorial works after any changes are made.

### DIFF
--- a/nameservice/happypath.sh
+++ b/nameservice/happypath.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+
+nscli query account $(nscli keys show jack -a) | jq ".value.coins[0]"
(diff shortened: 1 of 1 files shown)


nameservice/happypath.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+
+nscli query account $(nscli keys show jack -a) | jq ".value.coins[0]"
```
marbar3778: should change to test1 or change the `init.sh` script to reflect these names

okwme: made those updates in #279

marbar3778: ah fire!! haven't pulled master then

PR #302, opened 2020-04-22
nameservice tutorial edits

hxrts:
This is my first time going through the SDK docs so I'm trying to contribute feedback from this "naive read" in hopes to improve the docs UX a bit.

### DIFF
--- a/nameservice/tutorial/02-app-init.md
+++ b/nameservice/tutorial/02-app-init.md
@@ -13,7 +13,7 @@ scaffold app lvl-1 [user] [repo] [flags]
 cd [repo]
 ```
 
-Inside the repo you just created you'll see three directories: `app` which holds the definition of the main state machine runtime, `cmd` with commands for interacting with the runtime, and `x` that contains the application-specific state machine subset and message/transaction processing logic that will be composed with the main runtime to form the complete state machine.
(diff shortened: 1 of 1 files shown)


nameservice/tutorial/02-app-init.md
```diff
@@ -4,25 +4,26 @@ order: 02
...
-You can fill in `user` with your Github username and `repo` with the name of the repo you are creating.
+Run the following command, where `user` is your Github username and `repo` is the name of the repo where you'll be creating your application, `nameservice-tutorial` for example. For now you don't need any flags, but scaffold allows you to choose extra configuration options at this step if you like.
 ```bash
 scaffold app lvl-1 [user] [repo] [flags]
 
 cd [repo]
 ```
 
-In `app.go` it is defined what the application does when it receives a transaction. But first, it needs to be able to receive transactions in the correct order. This is the role of the [Tendermint consensus engine](https://github.com/tendermint/tendermint).
+Inside the repo you just created you'll see three directories: `app` which holds the definition of the main state machine runtime, `cmd` with commands for interacting with the runtime, and `x` that contains the application-specific state machine subset and message/transaction processing logic that will be composed with the main runtime to form the complete state machine.
```
okwme, line 16:
```suggestion
Inside the repo you just created you'll see three directories: `app` which holds the definition of the main state machine runtime, `cmd` with commands for interacting with the runtime, and `x` that contains the application-specific modules with message/transaction processing logic that will be combined with the main runtime to form the complete state machine.
```

hxrts:
this kind of stalled but there are some helpful edits on the first page. maybe worth merging anyway?

PR #310, opened 2020-05-11
closer to what current scaffold builds

okwme:


scavenge/go.mod
```diff
@@ -1,19 +1,17 @@
-module github.com/cosmos/sdk-tutorials/scavenge
+module github.com/okwme/scavenge
```
marbar3778, line 1:
```suggestion
module github.com/cosmos/sdk-tutorials/scavenge
```

will have to remove scavenge below

okwme: > will have to remove scavenge below

What do you mean by this actually?

okwme: remove from `go.mod`? (just did)

PR #328, opened 2020-06-07
Fix nameservice tutorial

nkmr-jp:
Fixed old parts of the documentation to match the actual code.


nkmr-jp:
There are also differences with the scaffold-generated code.
I've created a separate issue #329

PR #376, opened 2020-08-18
remove hellochain

marbar3778:
😢


okwme:
@hxrts no this is han's tutorial that came up with an alternative app structure to try to limit the complexity under the hood. Great place for brand new people to try out some very basic stuff but ultimately goes against the primary design recommendations.

sad to see it go 😔

PR #397, opened 2020-09-22
Migrate scavenge tutorial to use `starport` instead of `scaffold`

lukitsbrian:


lukitsbrian:
Currently, the content of this updated tutorial is consistent with the existing one, with a few changes (updating `module.go`, `main.go`, and `app.go` doesn't seem to be necessary), and overall it uses `starport` instead of `scaffold`.

There's another [branch](https://github.com/cosmos/sdk-tutorials/tree/scavenge-tut-ui) with a "Bonus REST + UI" portion of the tutorial (whose code already exists here) but I think we can push this out first since people are using `starport` for the hackathons and want to use the tutorials as reference.

.vuepress/config.js
```diff
@@ -40,9 +40,14 @@ module.exports = {
...
+            // {
+            //   title: "Scavenge",
+            //   path: "/scavenge/tutorial/",
+            //   directory: true,
+            // }, 
             {
-              title: "Scavenge",
-              path: "/scavenge/tutorial/",
+              title: "Starport Scavenge",
+              path: "/starport-scavenge/tutorial/",
```
marbar3778, line 50: Why not replace the main scavenge tutorial?

lukitsbrian: Good catch, thanks! I've removed the old tutorial

okwme:
From our last call it seemed like it would be a good idea to add back `init.sh` and `happypath.sh` and then when giving instructions for `starport serve` explaining what was happening and why. This would also be a chance to point out the `init.sh` file and say describe the more common method of running init every time and that `starport serve` is a quickly changing command with more and more advanced features (maybe a chance to link to any starport docs that may already exist).

.vuepress/config.js
marbar3778: thank you!!, should `starport-` be removed as well?

marbar3778:
This broke links @lukitsbrian @fadeev why does starport need to be in the name?

fadeev:
I don't think it has to be.

marbar3778:
welp looks like all the links are broken, because the folder structure change. Would have been nice to not do this or provide pages that when a user goes to the old pages it can point them to the right place, but i guess the cosmos community is used to broken links.

okwme:
I agree, could we change this back to `scavenge` instead of `starport-scavenge` to preserve links @lukitsbrian ?
Seems unnecessary unless there's another reason i'm not seeing

PR #400, opened 2020-09-28
Fix import statement in docs

rzadp:
The docs suggest adding import statement which is a markdown link.


marbar3778 (commented):
thank you

marbar3778 (approved):

PR #401, opened 2020-09-30
Migrate nameservice tutorial to use `starport`

lukitsbrian:


alessio:
Looks good to me (subscribed @okwme and @marbar3778 for visibility)

marbar3778:
Why is it named starport-nameservice? I would prefer to keep it nameservice to avoid breaking links.

alessio:
Good point @marbar3778 - @lukitsbrian mind updating the name?

PR #406, opened 2020-10-06
Add voter app code

lukitsbrian:
Add code for voter app


marbar3778 (changes requested):
what is the reasoning to have starport in the name? I feel like we can easily do without it

scavenge/scavenge/type.diff
```diff
@@ -1,18 +1,18 @@
-diff --git a/starport-scavenge/scavenge/vue/src/store/app.js b/starport-scavenge/scavenge/vue/src/store/app.js
+diff --git a/scavenge/scavenge/vue/src/store/app.js b/scavenge/scavenge/vue/src/store/app.js
```
marbar3778, line 1: never seen this file before. do we need it?

marbar3778 (approved):
utACK

scavenge/scavenge/type.diff
lukitsbrian: Removed it!

PR #454, opened 2020-12-07
Update nameservice 03-types with `whois`

Tosch110:
Make it clear to the user that `whois` should be added with the `starport type` command

### DIFF
--- a/nameservice/tutorial/03-types.md
+++ b/nameservice/tutorial/03-types.md
@@ -8,7 +8,7 @@ First thing we're going to do is create a type in the `/x/nameservice` folder wi
 
 
 ```bash
-starport type name value price
+starport type whois value price
 ```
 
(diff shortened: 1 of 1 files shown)


okwme (approved):

nameservice/tutorial/03-types.md
```diff
@@ -8,7 +8,7 @@ First thing we're going to do is create a type in the `/x/nameservice` folder wi
 
 
 ```bash
-starport type name value price
+starport type whois value price
 ```
 
 Currently, we're only passing two values when scaffolding the `whois` type, because there are additional fields that we will be replacing. In all uses of `whois`, we'll be removing the auto-generated `ID` field, and replacing the `Creator` field with `Owner` to reflect ownership of the name.
```
okwme, line 14:
```suggestion
Currently, we're only passing two values when scaffolding the `whois` type, because there are additional fields that we will be replacing. In all uses of `whois`, we'll be removing the auto-generated `ID` field and replacing the `Creator` field with `Owner` to reflect ownership of the name.
```

okwme (approved):

PR #459, opened 2020-12-08
Allow set-env

lukitsbrian:


.github/workflows/blog.yml
```diff
@@ -15,6 +15,8 @@ jobs:
         echo "::add-path::$(go env GOPATH)/bin"
         echo "::set-env name=GO111MODULE::"on""
       shell: bash
+      env:
```
marbar3778, line 18: by bumping setup-go to v2 you dont need to set these env variables.

marbar3778:
updated ci here: https://github.com/cosmos/sdk-tutorials/pull/463

PR #470, opened 2020-12-10
Starport nameservice tutorial error

leonvanderhaeghen:
found an error at https://tutorials.cosmos.network/nameservice/tutorial/08-set-name.html and https://tutorials.cosmos.network/nameservice/tutorial/09-buy-name.html .
in both files the GetSigners looks like this (msg.owner changes but syntax stays the same)

// GetSigners defines whose signature is required
func (msg MsgBuyName) GetSigners() []sdk.AccAddress {
	return []sdk.AccAddress{msg.Owner}
}

but it should be like this
// GetSigners defines whose signature is required
func (msg MsgBuyName) GetSigners() []sdk.AccAddress {
	return []sdk.AccAddress{sdk.AccAddress(msg.Buyer)}
}


lukitsbrian:
@leonvanderhaeghen Seems like you may have checked out an old branch (see #401). Could you open a new PR that directly addresses the fix? Your branch could be called `leonvanderhaeghen/starport-nameservice`.

PR #472, opened 2020-12-11
rectify error statements in genesis

Raneet10:
Previously in `nameservice/x/nameservice/types/genesis.go` : 

```
if record.Owner == nil {
			return fmt.Errorf("invalid WhoisRecord: Value: %s. Error: Missing Owner", record.Value)
		}
		if record.Value == "" {
			return fmt.Errorf("invalid WhoisRecord: Owner: %s. Error: Missing Value", record.Owner)
		}
		if record.Price == nil {
			return fmt.Errorf("invalid WhoisRecord: Value: %s. Error: Missing Price", record.Value)
		}
```


okwme (approved):
nice catch!

PR #479, opened 2020-12-18
Update pofe tutorial for Starport v0.13

fadeev:
We've updated templates and `pofe` tutorial will no longer work with the upcoming Starport v0.13. This PR should be merged once v0.13 is released next week.


Tosch110 (approved):

Tosch110 (changes requested):
The functions `GetClaimCount` and `SetClaimCount` that are used in the `CreateClaim` function seem to be not available.

fadeev:
No longer needed, because

https://github.com/cosmos/sdk-tutorials/pull/501

PR #483, opened 2021-01-05
Lowercase scavengeD and scavengeCLI for compatibility with linux

Tosch110:
As noted on Discord, uppercase D or CLI on the scavenge name works fine for mac, but not on linux:
https://discord.com/channels/669268347736686612/737461683588431924/795011101175316522

Therefore proposing to use all lowercase.


fadeev:
@Tosch110 We may also need to rename dirs inside the `cmd`.

Tosch110:
Thanks @fadeev - I did not notice GitHub not picking up these changes. i had to config github to `git config core.ignorecase false` - just for reference 👍 

Fixed

fadeev (approved):
`scavenge` works with starport v0.0.10 on Gitpod 👍

PR #490, opened 2021-01-07
Update Blog tutorial

### LINKED ISSUE #447: Blog tutorial is broken
the step to create the `MsgCreatePost.go` file is outdated. That file is already created.
```
Let’s define NewMsgCreatePost in a new file you should create as x/blog/types/MsgCreatePost.go.
```

`MsgCreateComment` is never defined but included in all of the instructions

a lot of the instructions are code snippets but they include the imports at the top. This makes it look like they are complete files and encourages a user to incorrectly copy the block and paste it in as the entire file. I'd suggest adding elipses to denote that tehre is code between the import statements, removing them, or setting them in separate code boxes.

`		msg := types.NewMsgCreatePost(addr, req.Title)` has too few arguments inside of the generated `/x/blog/client/rest/tx.go` file.

Tosch110:
Closes #447 

Updates Starport to v0.13.1


fadeev:
@Tosch110 great work! 🔥

blog/tutorial/01-index.md
```diff
@@ -186,15 +194,15 @@ func (msg MsgCreatePost) ValidateBasic() error {
...
 Going back to `GetCmdCreatePost` in `x/blog/client/cli/tx.go`, you'll see `MsgCreatePost` being created and broadcast with `GenerateOrBroadcastMsgs`.
 
 After being broadcast, the messages are processed by an important part of the application, called [**handlers**](https://docs.cosmos.network/master/building-modules/handler.html).
 
 ## x/blog/handler.go
 
-Begin by importing your new blog types that we created:
+Begin by checking if the new blog types that we created is imported:
 
 ```go
```
fadeev, line 207:
```go
import (
  // Existing imports...
  "github.com/example/blog/x/blog/types"
)
```

@Tosch110 can you please double-check, but I believe the default scaffold already imports `types` package, so if a user includes this import (and their editor isn't smart), Starport throws an error.

fadeev (dismissed):

blog/tutorial/01-index.md
Tosch110: @fadeev this is correct, that's why I changed the wording above the sentence. Do you think we should rather erase this part then?

fadeev: You're right, my bad 🙂

fadeev:
I'm getting an error with `x/blog/keeper/querier.go`. It seems like we're importing `"github.com/cosmos/cosmos-sdk/codec"`, but not using `codec` anywhere.

fadeev:
Thanks for the quick changes! I believe this is the last thing I have a problem with. In the "List posts with HTTP" section in `x/blog/client/rest/rest.go` imports are unused:

```
import (
  // Existing imports...
  "fmt"
  "net/http"
  "github.com/cosmos/cosmos-sdk/types/rest"
)
```

The UI also has changed, so in web UI section the only change we need is to make:

```
    <div class="sp-container">
      <sp-sign-in />
      <sp-bank-balances />
      <sp-token-send />
      <!-- this line is used by starport scaffolding # 4 -->
      <sp-type-form type="posts" :fields="['title', 'body']"/>
    </div>
```

We can add the following image as the illustration:

Screenshot 2021-01-08 at 00 34 30

We may also need to remove the section on how to do it manually, since it's using `@tendermint/sig`, and it's no longer maintained.

Tosch110:
Updated as requested, removed the manual `frontend` explanation and code from the repo.

I've updated the `sdk-tutorials/blog/blog` deployment from previous versions to boostrapped with Starport v0.13.1

fadeev (approved):
Looks good to me! 👍

PR #494, opened 2021-01-10
fix invalid starport command in blog tutorial

mimoo:


okwme:
you might need to update your starport version. This flag was added because there are a number of versions of starport floating around at this point and in order to make sure tutorials work they need to be explicit about the version. Maybe instead we should add a comment at the beginning to ensure that starport is updated and that if it's not you'll run into errors?

mimoo:
I just installed the last version using the command in the tutorial : o

okwme:
uh oh that's not good, was that flag removed @fadeev ?

fadeev:
Every tutorial now includes a version of Starport that is recommended for this particular tutorial. `blog` has been updated to `v0.13.1`, see "Requirements":

```
curl https://i.jpillora.com/tendermint/starport@v0.13.1! | bash
```

"Getting started":

```
starport app github.com/example/blog --sdk-version launchpad
```

Trying it out:

Screenshot 2021-01-11 at 10 43 24

mimoo:
oh indeed, it looks like I was on an old version, even though I installed it this week end. I got the command to install it from the docs, but not sure where anymore : o

PR #524, opened 2021-02-01
Updating blog to use starport v0.13.2:

seantking:
When using Starport v0.13.2 some parts of the tutorial are incorrect. For example:

`client.ReadTxCommandFlags` is no longer used due to this commit -> https://github.com/cosmos/cosmos-sdk/commit/a3ae6e3704e66c966f6dab6c71646738dd2be0ed

Also:

`return clientCtx.PrintOutput(res)` is now replaced with -> `return clientCtx.PrintProto(res)`

* Updating cli client commands to use new functions
* Adding Error fix suggestion
* Generating new finished blog project to use new starport version

I also added the comment type to the blog project as it's technically the final step in the tutorial (thus all the newly generated code added).


marbar3778 (approved):
Looks good! Crazy that a minor version bump causes this many changes

PR #525, opened 2021-02-01
Update nameservice

### LINKED ISSUE #450: Nameservice out of date
`key.go` missing `NamePrefix` that is there now with starport

Keeper makes reference to generated `WhoIs` keeper that doesn't exist. Was there a step for using starport with whois?
unclear if the expanded keeper file is supposed to overwrite the contents of `keeper/keeper.go` or go in a new file. if it overwrites then there is missing `NewKeeper`

going to stp here, seems like it's also outdated and would recommend same path forward as scavenge tutorial

Tosch110:
Update Nameservice tutorial to Starport v0.13.1

Closes #450


nameservice/nameservice/config.yml
```diff
@@ -1,9 +1,9 @@
 version: 1
 accounts:
-  - name: jack
```
lukitsbrian, line 3: In previous versions of the tutorial we use jack/alice as opposed to user1/user2, so [21-run-rest](http://localhost:8080/nameservice/tutorial/21-run-rest.html) should also be updated.

lukitsbrian (approved):

nameservice/nameservice/config.yml
Tosch110: Thanks. Addressed 👍

PR #540, opened 2021-02-20
Update version of Voter to Starport v0.14.0

fadeev:
Tested tutorial with the latest Starport v0.14.0, works just fine.


fadeev:
@Tosch110 can you review this PR, please 🙏 I've got a suspicion that Gitpod has changed something, and links to Starport v0.13.1 open an environment, where the binary is not installed. v0.13.2 and v0.14.0 work fine, so we may need to update to those versions.

Tosch110 (approved):

PR #623, opened 2021-04-29
Remove v0.39 tutorials from the navbar

toschdev:
Developers using the tutorials should work with Stargate version v0.40 plus of the Cosmos SDK.

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -28,7 +28,7 @@ module.exports = {
       auto: false,
       nav: [
         {
-          title: "v0.40+ Stargate Tutorials",
+          title: "Tutorials",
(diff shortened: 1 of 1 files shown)


fadeev:
To make it even cleaner, since we no longer have Launchpad tutorials, we can drop "V0.40+ STARGATE TUTORIALS" (leave it blank) and rename "MIGRATE COSMOS SDK V0.39 TO V0.40" to "Migrate to Stargate".

toschdev:
Thanks @fadeev 
I have renamed the migration title. But leaving blank is not an option, the whole section then disappears in the navbar.

fadeev (approved):

PR #667, opened 2021-05-20
tutorials: add description metadata to index and intro files

barriebyron:
Add description metadata to the front matter of all index or intro files so that a useful summary shows on links and search results.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,3 +19,6 @@ dist/
 # IDE
 .idea/
 *.iml
+
+# vscode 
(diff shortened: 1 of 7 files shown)


barriebyron:
@fadeev I came across an instance recently where a tutorial search result showed a very long string of the content instead of an abstract... maybe this description metadata will help with that UX?

fadeev (approved):
@barriebyron yes, descriptions should help. Thanks for adding them!

PR #741, opened 2021-06-29
Minor update

alijnmerchant21:
- Remove relayer file (relayer file is within .starport)
- Add tip - check relayer terminal window. (To confirm if tx went through)
- Add number of tokens to be transferred.

### DIFF
--- a/liquidity-module/index.md
+++ b/liquidity-module/index.md
@@ -237,7 +237,7 @@ After your transaction is complete, check your balance on the Gravity DEX testne
 
 Make sure to replace `cosmosxxxxx` with your address.
 
-**Tip** Sometimes transactions don't go through in first try. Make sure you check your terminal window created by your relayer. It should have the following output.
+**Tip** Sometimes transactions don't go through on the first try. Make sure you check the terminal window that shows the relayer process and verify that you see output similar to the following output:
 
  ```markdown
  Relay 1 packets from myblockchain => cosmoshub-testnet
--- a/liquidity-module/index.md
+++ b/liquidity-module/index.md
@@ -129,12 +129,6 @@ A relayer is software to connect two blockchains. Configure the relayer with you
 
 If you previously used the relayer, follow these steps to remove exiting relayer and Starport configurations:
 
-- Remove your existing relayer config:
-    
-    ```bash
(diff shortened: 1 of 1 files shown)


liquidity-module/index.md
```diff
@@ -129,12 +129,6 @@ A relayer is software to connect two blockchains. Configure the relayer with you
 
 If you previously used the relayer, follow these steps to remove exiting relayer and Starport configurations:
 
-- Remove your existing relayer config:
```
barriebyron, line 132: we don't need to remove the existing relayer config now? if you remove this step then remove the mention in line 130 too

liquidity-module/index.md
```diff
@@ -236,13 +230,22 @@ At your local terminal, enter the IBC module command to transfer your token to t
...
 ```bash
-myblockchaind tx ibc-transfer transfer transfer channel-0 cosmosxxxxx "500mytoken" --from username
+myblockchaind tx ibc-transfer transfer transfer channel-0 cosmosxxxxx "15000000mytoken" --from username
 ```
 
 After your transaction is complete, check your balance on the Gravity DEX testnet to confirm your token transfer.
 
 Make sure to replace `cosmosxxxxx` with your address.
 
+**Tip** Sometimes transactions don't go through in first try. Make sure you check your terminal window created by your relayer. It should have the following output.
```
barriebyron, line 240:
```suggestion
**Tip** Sometimes transactions don't go through on the first try. Make sure you check the terminal window that shows the relayer process and verify that you see output similar to the following output:
```

barriebyron: hi @alijnmerchant21 ! which terminal window was "created by your relayer"?

barriebyron (approved):
I left a few questions and comments, but approving so you can merge after you address the feedback, it's nice to see your PR here! good job

liquidity-module/index.md
alijnmerchant21: We have to remove the relayer config. But the relayer file exist within .starport directory. So, removing .starport removes both starport and relayer files.

liquidity-module/index.md
alijnmerchant21: I have attached the o/p as a markdown comment

Issue #738, opened 2021-06-28
Interchain Exchange: return to original token


Kmortyk:
**Tutorial URL**
https://tutorials.cosmos.network/interchain-exchange/tutorial/09-walkthrough.html#exchange-a-token-and-return-to-original-token

**Description**
Hello, I cannot create an order pair for ibc vouchers. The package arriving at another node is incorrect.

**To reproduce**
1. Exchange a pair of tokens using `send-createPair, send-sellOrder and send-buyOrder`.
2. Try to create order pair for vouchers with   
`interchanged tx ibcdex send-createPair ibcdex channel-0 ibc/50D70B7748FB8AA69F09114EC9E5615C39E07381FE80E628A1AF63A6F5C79833 ibc/99678A10AF684E33E88959727F2455AE42CCC64CD76ECFA9691E1B5A32342D33 --from alice --chain-id mars --home ~/.mars
`.

![image](https://user-images.githubusercontent.com/16573297/123606464-96b7dc00-d805-11eb-81f5-de171b89d141.png)

**Environment**
- `starport version v0.16.2 linux/amd64 -build date: 2021-06-22T05:43:46Z`
- `go version go1.16.2 linux/amd64`
- `Linux kmortyk-desktop 5.4.0-77-generic #86~18.04.1-Ubuntu SMP Fri Jun 18 01:23:22 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux`

lubtd:
Thanks for catching this.

It seems the problem is linked to the denom itself since we have the same issue if you directly try to create the pair first

lubtd:
We can for now fix this by limiting the length of the IBC voucher denom https://github.com/tendermint/interchange/blob/6bae1ffc3e69365cce64850a4f812446463dd56f/x/ibcdex/keeper/denom.go#L23 since the issue is related to the length of the denom

PR #745, opened 2021-06-30
fix: ibc voucher denom issue

lubtd:
Reduce the generated IBC voucher denom as it makes the IBC packet invalid (for a reason that is still unknown and that should be investigated)


toschdev:
I don't really understand. What error do you see with the regular denom length?

lubtd:
> I don't really understand. What error do you see with the regular denom length?

https://github.com/cosmos/sdk-tutorials/issues/738 if the length of the denom is too long when creating a pair (which is the case with IBC vouchers) the packet fir pair creation can't be processed

toschdev (approved):

PR #757, opened 2021-07-11
Change to correct port numbers

lilamila:


toschdev:
Thanks a lot for your Pull Request @lilamila 

Although we update the output coherent with the indicated version at the beginning of the tutorial.

PR #764, opened 2021-07-15
Update blog tutorial to Starport v0.17

toschdev:
Update blog tutorial to Starport v0.17

For easier review, text and code split in two PRs:
text: https://github.com/cosmos/sdk-tutorials/pull/764
code: https://github.com/cosmos/sdk-tutorials/pull/765

PR #765, opened 2021-07-15
Update blog tutorial code to Starport v0.17

toschdev:
Update blog tutorial code to Starport v0.17

This is the code implementation of the blog tutorial. We cannot make this update more granular because it is the result of using Starport as described in the application. And many things change with Starport with each release.
It is not important to review the files. But that:

- The application can get installed using the Makefile
- The application runs with starport chain serve
- The application does have the commands described in the tutorial.

For easier review, text and code have been split into two PRs:

- text: https://github.com/cosmos/sdk-tutorials/pull/764
- code: https://github.com/cosmos/sdk-tutorials/pull/765


barriebyron (changes requested):
oooh I'm still confused. Do we really want 766 files changed in this PR? just explain and reselect me as reviewer if that's expected. For changes this large, wdyt about adding those details in the description? (I just get uncomfortable when I see this many files being changed)
help me understand, Barrie

toschdev:
Thanks Barrie. This is the code implementation of the blog tutorial. Sadly we cannot make it more granular because it is the result of using Starport as described in the application. And many things change with Starport with each release. 
It is not important to review the files. But that:

- The application can get installed using the Makefile
- The application runs with `starport chain serve`
- The application does have the commands described in the tutorial.

barriebyron (approved):
thanks for explaining

PR #766, opened 2021-07-15
Update voter tutorial to Starport v0.17

toschdev:
Update voter tutorial to Starport v0.17

For easier review, text and code split in two PRs:
text: #766
code: #767

PR #767, opened 2021-07-15
Update voter tutorial code to Starport v0.17

toschdev:
Update the code for voter tutorial to Starport v0.17

For easier review, text and code split in two PRs:
text: #766
code: #767


barriebyron (changes requested):
wait! do we really want to see 710 files changed for this tutorial? We're changing the code but is this large number of files as expected? eek @toschdev

toschdev:
Thanks Barrie. This is the code implementation of the voter tutorial. Sadly we cannot make it more granular because it is the result of using Starport as described in the application. And many things change with Starport with each release.
It is not important to review the files. But that:

- The application can get installed using the Makefile
- The application runs with starport chain serve
- The application does have the commands described in the tutorial.

barriebyron (approved):
I see how this works now!

PR #768, opened 2021-07-16
Remove legacy tutorials

toschdev:
Legacy tutorials now create a confusion in our repository. The initial goal was to provide support to tutorials for Cosmos v0.39 and prior. 
We now want to target Cosmos SDK v0.40 and later, therefore we do not maintain the tutorials with an older version anymore.

Therefore we remove legacy tutorials. A release has been created prior to removing them, to have a state to point to where to find the legacy tutorials: https://github.com/cosmos/sdk-tutorials/releases/tag/launchpad-backup


barriebyron (approved):
thanks for doing this!

PR #778, opened 2021-07-28
closed moving focus

barriebyron:
Starts to fulfill the plan to provide a tutorial templates structure for our tutorials. We want as many contributors as we can get, and providing structure and examples are a good starting point. 

Use these templates to help you create articles for the Cosmos community.

Choose the template that best fits the kind of content you're creating

### DIFF
--- a/templates/README.md
+++ b/templates/README.md
@@ -1,0 +1,14 @@
+# cosmos-article-templates — Cosmos and Starport Community Article Templates
+
+Use these templates along with our style guide to help you create articles for the Cosmos community. 
+
+Choose the template that best fits the kind of content you're creating:
+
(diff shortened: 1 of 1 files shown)


barriebyron:
@fadeev and @toschdev 
what is our branding for tutorials, moving forward. Since we're moving our tutorials to tendermint/starport I should probably move the templates folder and contents to that repo? and use Starport as the branding?
You'll see I waffle around, sometimes it's "Cosmos ecosystem" and sometimes its "Tendermint" and sometimes I use "The Tendermint Ecosystem Development Team"

I realize I am unclear, and if I am, then so are our users. Do we have a firm branding plan? Starport everything? or?

PR #809, opened 2021-09-10
B9lab platform updates 2 - Enable preview of B9lab content

coldice:
This PR adds a function which allows viewers to access the B9lab content modules and sections through the main menu on **preview deployments and when serving locally**, while keeping them hidden on the public site. This is implemented through a [javascript method](https://github.com/cosmos/sdk-tutorials/compare/b9lab-platform-updates-2?expand=1#diff-286e37e3164f39dd52b9f31491f5f2764cb93e31744df6ece7f85aef24fde671R221), which will only set the menu items visible if serving on:

* deploy-preview
* localhost
* 127.0.0.1

This functionality can be removed once the B9lab content is published.

screeen_pr_1

### DIFF
--- a/.vuepress/config.js
+++ b/.vuepress/config.js
@@ -121,7 +121,6 @@ module.exports = {
             },
           ],
         },
-        /* Start of B9lab Content - WIP
         {
           title: "B9lab content",
           children: [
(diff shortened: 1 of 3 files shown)


okwme (approved):
bravo!

PR #813, opened 2021-09-15
Update Starport install URL

fadeev:
The URL is now https://docs.starport.network/guide/install.html


soniasingla (approved):
thanks! 💪

fadeev:
@barriebyron we can, but the list of redirects will become messy over time. The best we can do is design an intuitive sidebar and a good not found page.

barriebyron:
@fadeev oooh a good not found page... will investigate.

toschdev (approved):

PR #814, opened 2021-09-15
fix:(liquidity-module/index.md) minor edits for style

barriebyron:
- small edits for cleanup
- use present tense
- provide the "replace with" after the example commands
- consistency

### DIFF
--- a/liquidity-module/index.md
+++ b/liquidity-module/index.md
@@ -179,13 +179,13 @@ Connect the chains:
 starport relayer connect
 ```
 
-It will start to connect your two blockchains, you will see an output while it does
+As your two blockchains start to connect, you see output like:
 
 ```bash
 ◣ Linking paths between chains... 
 ```
 
-When successful, your output will be
+When successful, your output shows:
 
 ```bash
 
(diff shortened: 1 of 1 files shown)


liquidity-module/index.md
```diff
@@ -303,50 +308,51 @@ Add `username`s mnemonic passphrase and hit enter. You will see something simila
...
-[https://api.testnet.cosmos.network/tendermint/liquidity/v1beta1/pools](https://api.testnet.cosmos.network/tendermint/liquidity/v1beta1/pools)
+- Visit [https://api.testnet.cosmos.network/tendermint/liquidity/v1beta1/pools](https://api.testnet.cosmos.network/tendermint/liquidity/v1beta1/pools)
 
 
 ## Swap Token
 
 You are ready to swap tokens! You now have uphoton token in your account and want to swap for the new IBC coin:
 
 ```bash
-gaiad tx liquidity swap 1 1 100000uphoton ibc/longibchash 0.1 0.003 --from username --chain-id cosmoshub-testnet --gas-prices "0.025stake" --node https://rpc.testnet.cosmos.network:443 -y
```
toschdev, line 324: Why not? 😅

toschdev (approved):

toschdev:
Thanks for the edits and improvements @barriebyron 🙏

soniasingla (approved):
lgtm 🎉

liquidity-module/index.md
barriebyron: the theory behind optional steps in tutorials is that we want to provide the core experience for our users, no need to add in optional steps since it does not add to learning and just adds one more teensy bit of noise to explain

alijnmerchant21: Agreed @barriebyron But I think we should also find a way to encourage people to look at different flags and what it does. Any suggestions how can we do that?

barriebyron: I always want to know what the flags do, too! 
Maybe an optional step? 
where are the flags documented? In the help and Starport CLI reference https://docs.starport.network/cli/ we really don't expose what the flags are. Teach me!

PR #823, opened 2021-09-21
docs: Module 3 - Image sketches - (B9lab updates 2-8)

coldice:
Docs content update for Module 3 (Main Concepts), **Image sketches** for multiple sections - part of the [b9lab-content](https://github.com/cosmos/sdk-tutorials/tree/master/b9lab-content).

**Note: ** These are just the sketches for the images (pre-production)

### Content state

**✔️ Completed**

### Technical review

- [x] requested
- [x] completed

### Language review

- [ ] requested
- [ ] completed


cmwaters (commented):
Hey, Noam asked me to have a look at these and so I have put together a few comments about the images:

1. `architecture_overview.png` If what ABCI stands for hasn't been mentioned yet, I'd be more explicit and call this Figure: Application Blockchain Interface (ABCI) Architecture. Also I don't know if you want to symbolize what data actually get's passed in these calls. Basically Tendermint is sending an agreed set of transactions to the application and the application is returning validator set updates and the hash of the new state (I'm simplifying obviously but I think the other information that gets sent is irrelevant).
2. `delegated-reward.png` This seems all good. Just a minor nit: it might be better if Validator 1 had 100 total weight instead of 120. People might be able to more logically see that Validator 2 mints twice as many blocks because they have twice as many ATOMs at stake as Validator 1.
3. `delayed-execution-model.png` I think this is still quite confusing and it might be more comprehensible for people if we remove all the stuff about validators and just focus on the relationship between blocks and application state. Hopefully before seeing this, people know that at a high level tendermint can be seen as continually producing batches of agreed upon transactions for an application to execute on and thus update state. The application is then responsible for returning the hash (or some representation) of that state to be put in the following block. It is therefore only in the following block (n + 1) that agreement of state at height n is made.
4. `minting-validation.png` I'm not 100% sure of the context behind this image but if we're talking about Tendermint finalization, that depends on 2/3+ voting on a block not 51%.
5. `next-validator-pick.png` This is a nice way of depicting a weighted randomized proposer selection algorithm. Tendermint isn't however randomized, it's round-robin. If all validators had even weighting, then the proposer of one height would move to the back of the line for the next height. Given different weightings a validator can sometimes find themselves at the front of the queue quicker or slower than other validators. The point is that if the validator set remains the same then the exact order of proposers are known.

xavierlepretre:
@cmwaters @clevinson Some updates from my end.

* `architecture_overview`: The reader has been introduced to the ABCI acronym beforehand. Here, we don't detail the type of information in ABCI. Although it is worth adding another diagram that zooms onto the ABCI arrows: `abci-messaging`. The reason there is another node is to show what parts communicates with the outside world, even if the word networking already embeds that concept.
* `delegated-reward`: Good points. I made it such that validator 1 has 100 coins and validator 2 has 200. The rewards are then unbalanced over 8 blocks: validator 1 gets more. The point is emphasised with a note about delegators 2 and 4, which both delegate the same amount. Also the blocks are picked in a round-robin fashion. Also, I renamed ATOM to the more generic 💰. The eventual depiction will be not ATOM but something better than this emoji. Also added a bubble about 0% commission.
* `delayed-execution-model`: I suggest to have 2 diagrams. The first very simple, to drive the point home. The second one similar to the old one, with disambiguation about the other validators. Depending on how it makes sense we could have those 2 (or more) overlay each other in a proto-animation.
* `minting-validation`: The point is to show that, not only the minter may change for every block, but also it is not necessary to have all validators validate all blocks, and that those that validate may differ from one block to the next. I picked different validators on the second block to reduce confusion and quietly show a round-robin.
* `next-validator-pick`: Point taken, wheel of fortune is out. I did not understand the Stable Set visual. Would the following visualisation make sense? A deck of colourful cards where each validator is represented by one colour. There may be many cards of a given colour: 1 validator has many cards. The richer the validator, the more cards of its colour in the deck. The deck is shuffled every time there is a change in the validators and their weights. The card corners are seen, that's the predictability. After you have taken one from the top you send it to the bottom: round-robin.

Let me make a commit.

cmwaters:
- The `architecture-overview` Looks good to me. Only thing that comes to mind is to change `Tendermint/ABCI architecture` to `Blockchain Node Architecture`. I feel like the figure is more showing how the components of a node come together (not just of Tendermint or ABCI). 
- For `delayed-execution-simple.jpg`, I was thinking something more like this:
delayed-execution
- `delegated-reward`: 👍 
- `minting-validation`: 👍  only things here is that we usually use the term proposer instead of minter and that the proposer (unless someone modifies the code) always votes for their own proposal. 
- `next-validator-pick`: I think your analogy kind of makes sense. I can quickly try explain how the proposer priority works. I found the Stable Set visual also confusing. Imagine you had three validators:

| Validators | val1 | val2 | val3 |
|-----------|-----|-----|-----|
| Voting Power | 300 | 200 | 100 | 
| Priority | 300 | 200 | 100 | 

At height 1, the priority starts equal to the voting power. val1 has the most voting power so they propose first. After that height we calculate the new priorities by first adding the voting power of each respective validator to the priority and then subtracting the total voting power (600) from the validator that was the proposer in the last height (val1). The new priorities looks as follows:
| Validators | val1 | val2 | val3 |
|-----------|-----|-----|-----|
| Voting Power | 300 | 200 | 100 | 
| Priority | 0 (300 + 300 - 600) | 400 (200 + 200) | 200 (100 + 100) | 

Now `val2` has the most priority and they propose.

xavierlepretre:
@cmwaters 

* `architecture-overview` good point. Changed.
* `delayed-execution-simple`: Good point. I updated both diagrams.
* `minting-validation`: Updated. Also with the validators included in the 66%, which was an oversight from my end.
* `next-validator-pick`: Easier to understand. I'll sleep on it. It could well be just a table with colourful cells.

coldice:
Merged in b9lab-content-updates-rc1 (no diff to this branch)

PR #834, opened 2021-10-08
minor

rob-Hitchens:
`cosmos-sdk` literal


barriebyron (approved):
every little bit of precision helps, thanks @rob-Hitchens

PR #838, opened 2021-10-08
bootstrap validator clarification

rob-Hitchens:


coldice:
already merged in via review updates

PR #839, opened 2021-10-09
Update index.md with error mesaage from Ubuntu.

hannydevelop:
Ubuntu 20.04 gives an error when installing Starport


nooomski:
Closing this as duplicate exists on Starport page already

PR #849, opened 2021-11-02
docs: Module 3 Multistore Keepers - code sample integration (B9lab updates 2-3-a)

coldice:
Docs content update for Module 3, Section _Multistore Keepers_ - part of the [b9lab-content](https://github.com/cosmos/sdk-tutorials/tree/master/b9lab-content).

This PR adds code samples to the content. The code shown here is taken from the [code example repo](https://github.com/cosmos/b9-checkers-academy-draft/) and has been hardly adjusted. The repo has been reviewed so no need to review it here. What we are looking for is to make sure that the introductory texts are not misleading or even downright wrong.

### Content state

**✔️ Completed**

### Technical review
- [x] requested
- [x] completed

### Language review
- [x] requested
- [x] completed


b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
 
 BaseApp holds an anteHandler as a parameter that is initialized in the application's constructor. The most widely used anteHandler is the auth module.
 
-More info: https://github.com/cosmos/cosmos-sdk/blob/master/docs/basics/gas-fees.md#antehandler + https://github.com/cosmos/cosmos-sdk/blob/master/docs/basics/gas-fees.md 
+More info: https://github.com/cosmos/cosmos-sdk/blob/master/docs/basics/gas-fees.md#antehandler + https://github.com/cosmos/cosmos-sdk/blob/master/docs/basics/gas-fees.md
```
nooomski, line 111: Don't think we need to have two links to the same article. Would also suggesting adding the antihandler doc on docs.cosmos.network

```suggestion
More info: https://docs.cosmos.network/master/modules/auth/03_antehandlers.html + https://github.com/cosmos/cosmos-sdk/blob/master/docs/basics/gas-fees.md#antehandler
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+
+<ExpansionPanel title="Show me some code for my checkers blockchain">
+
+In the [Accounts](./04-accounts) section, you were introduced to the elements of the stored game, but were left in the dark about where this game is stored. In light of what you learned above, let's remedy that.
```
nooomski, line 115: Style suggestion

```suggestion
In the [Accounts](./04-accounts) section, you were introduced to the elements of the stored game, but were left in the dark about where this game is stored. In light of what you learned above, let's fix that.
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+
+## Game Object in Storage
+
+You need to decide under what structure to store a game in storage. The Cosmos SDK partitions the global storage per module, with `checkers` being its own module. Therefore, you need to take care of how to store games in the checkers' module's own corner of the key / value pair storage.
```
nooomski, line 119:
```suggestion
You need to decide under what structure to store a game in storage. The Cosmos SDK partitions the global storage per module, with `checkers` being its own module. Therefore, you need to take care of how to store games in the checkers module's own corner of the key / value pair storage.
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+
+The first idea would be to attribute a unique id to a game and to store the game value at that id. However, for the sake of clarity and so as to be able to differentiate with other stored elements in the future, you ought to add a prefix to that id. In pseudo-Go-code, the storage structure would therefore look like this:
```
nooomski, line 121:
```suggestion
The first idea would be to attribute a unique id to a game and to store the game value at that id. However, for the sake of clarity and so as to be able to differentiate with other stored elements in the future, you ought to add a prefix to that id. In pseudo Go code, the storage structure would therefore look like this:
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+        k.cdc.MustUnmarshalBinaryBare(iterator.Value(), &storedGame)
+        list = append(list, storedGame)
+    }
+
+    return
+}
+```
+Notice the `MustMarshalBinaryBare` and `MustUnmarshalBinaryBare` functions in the `codec` above. They need to be instructed as to how to proceed with the marshaling. Fortunately, Protobuf took care of this for us. See the [previous section](./09-protobuf) for that.
+
+## Boilerplate Boilerplate Everywhere!
```
nooomski, line 209:
```suggestion
## Boilerplate, Boilerplate Everywhere!
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+    panic(err)
+}
+if deadline.Before(ctx.BlockTime()) {
+    // TODO
+}
+```
+
+## How To Expire Games
+
+Ok, but when do you verify that the game has expired? An interesting feature of an ABCI application is that you can do some action at the end of each block. Should you load all games and filter for those that have expired? No. That would be extremely expensive. Better keep a FIFO where fresh games are pushed back to the tail and therefore the head contains the next games to expire.
```
nooomski, line 386:
```suggestion
Ok, but when do you verify that the game has expired? An interesting feature of an ABCI application is that you can have it perform some actions at the end of each block. Should you load all games and filter for those that have expired? No. That would be extremely expensive. Better keep a FIFO where fresh games are pushed back to the tail so that the head contains the next games to expire.
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+    // TODO
+    return []abci.ValidatorUpdate{}
+}
+```
+This is where you write the necessary code, preferably in the keeper. For instance with:
+
+```go
+am.keeper.ForfeitExpiredGames(sdk.WrapSDKContext(ctx))
+```
+Those among you with a well-placed paranoia must be asking whether you can ensure that the execution of this `EndBlock` does not become prohibitively expensive. After all, the number of games to potentially expire is unbounded, which is a recipe for disaster in the blockchain world. Is there a situation or an attack vector that would make this is a possibility?
```
nooomski, line 430:
```suggestion
Those among you with a well-placed paranoia must be asking whether you can ensure that the execution of this `EndBlock` does not become prohibitively expensive. After all, the number of games to potentially expire is unbounded, which is a recipe for disaster in the blockchain world. Is there a situation or an attack vector that would make this a possibility? And what can we do to prevent it?
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+
+Fortunately, the timeout duration is fixed and the same for all games. This means that those `n` games that all expire in a given block have all been created or updated at roughly the same time. Or in effect, at roughly the same block height `h` give or take a margin of error `h-1` and `h+1`. These created and updated games are in limited number, as per the validators rules. So if by any chance, all games in blocks `h-1`, `h` and `h+1` expire now, then the `EndBlock` function would have to expire 3 times as many games as a block can handle. This is the worst case scenario, and still sounds manageable.
+
+As a corollary, you should be careful about letting the game creator pick a timeout duration. That would open an avenue for a malicious actor to, for instance, stagger game creations over a large number of blocks, with decreasing timeouts, so that they all expire at the same time.
```
nooomski, line 434: Keeping it simple :)
```suggestion
Because of this, you should be careful about letting the game creator pick a timeout duration. That would open an avenue for a malicious actor to, for instance, stagger game creations over a large number of blocks, with decreasing timeouts, so that they all expire at the same time.
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+}
+```
+This is where you write the necessary code, preferably in the keeper. For instance with:
+
+```go
+am.keeper.ForfeitExpiredGames(sdk.WrapSDKContext(ctx))
+```
+Those among you with a well-placed paranoia must be asking whether you can ensure that the execution of this `EndBlock` does not become prohibitively expensive. After all, the number of games to potentially expire is unbounded, which is a recipe for disaster in the blockchain world. Is there a situation or an attack vector that would make this is a possibility?
+
+Fortunately, the timeout duration is fixed and the same for all games. This means that those `n` games that all expire in a given block have all been created or updated at roughly the same time. Or in effect, at roughly the same block height `h` give or take a margin of error `h-1` and `h+1`. These created and updated games are in limited number, as per the validators rules. So if by any chance, all games in blocks `h-1`, `h` and `h+1` expire now, then the `EndBlock` function would have to expire 3 times as many games as a block can handle. This is the worst case scenario, and still sounds manageable.
```
nooomski, line 432: What exactly is meant with "as per the validators rules" ? @xavierlepretre

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+Then, in your `MsgCreateGame` handler, you add the line:
+
+```go
+func (k msgServer) CreateGame(goCtx context.Context, msg *types.MsgCreateGame) (*types.MsgCreateGameResponse, error) {
+    ...
+    ctx.GasMeter().ConsumeGas(types.CreateGameGas, "Create game")
+    ...
+}
+```
+Where and how much to charge for gas is an economics as well as a game theoretic concern. So you should think about what you want to incentivize or disincentivize, at which junctures, and charge accordingly.
```
nooomski, line 462:
```suggestion
Where and how much to charge for gas is an economical, as well as a game theoretical concern. So you should think about what you want to incentivize or disincentivize, at which junctures, and charge accordingly.
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
 
-Here we also introduce the idea of gas costs. Ratios proposal are to be adjusted so it makes sense compared to the base transaction costs:
+* Create game: costs 10. Conceptually, it should include the costs of closing a game. If that was not the case, the losing player would be incentivized to let the game hit its timeout.
+* Play move: costs 1. You could make it cost 0 when the player loses the game, in order to incentivize the player to conclude the game instead of letting it hit the time out.
+* Reject: costs 0 because you want to incentivize cleaning up the state. This transaction would still cost what Cosmos SDK bills for transactions.
 
-* `CreateGameTx`: costs 10. Conceptually, it should include the costs of closing a game. If that was not the case, the losing player would be incentivized to let the game hit its timeout.
-* `MoveTx`: costs 1. Perhaps make it cost 0 when the player loses the game, in order to incentivize the player to conclude the game instead of letting it hit the time out.
-* `RejectTx`: costs 0 because we want to incentivize cleaning up the state. This transaction would still cost what Cosmos SDK bills for transactions.
+So you define the cost:
```
nooomski, line 444: Do we need to say where to define these costs? Or is it still abstract / theoretical @xavierlepretre ?

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
+
+As a corollary, you should be careful about letting the game creator pick a timeout duration. That would open an avenue for a malicious actor to, for instance, stagger game creations over a large number of blocks, with decreasing timeouts, so that they all expire at the same time.
 
-* Remove all traces of the game when it is finished.
+## Gas costs
 
-Because of these requirements, we think about an appropriate storage structure.
+The keeper also makes it easy for you to charge gas to the players as you see fit. Let's propose some ratios, which would have to be adjusted so it makes sense compared to the base transaction costs:
```
nooomski, line 438: Added for clarity. Did I get this right @xavierlepretre ?

```suggestion
The keeper also makes it easy for you to charge gas to the players as you see fit. This is a gas fee that comes on top of the configured standard fee for transactions on your chain. Let's propose some ratios, which would have to be adjusted so it makes sense compared to the base transaction costs:
```

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -108,27 +108,357 @@ The AnteHandler is theoretically optional, but still a very important component
...
 
-Here we also introduce the idea of gas costs. Ratios proposal are to be adjusted so it makes sense compared to the base transaction costs:
+* Create game: costs 10. Conceptually, it should include the costs of closing a game. If that was not the case, the losing player would be incentivized to let the game hit its timeout.
+* Play move: costs 1. You could make it cost 0 when the player loses the game, in order to incentivize the player to conclude the game instead of letting it hit the time out.
+* Reject: costs 0 because you want to incentivize cleaning up the state. This transaction would still cost what Cosmos SDK bills for transactions.
```
nooomski, line 442:
```suggestion
* Reject: costs 0 because you want to incentivize cleaning up the state. This transaction would still cost what your chain is configured to charge for basic transactions.
```

nooomski (commented):
Really great, very in depth. A few things for @xavierlepretre to clarify, but nothing critical for launch except the TODO on line 49.

b9lab-content/3-main-concepts/10-multistore-keepers.md
xavierlepretre: When a validator builds the list of transactions that go into the next block it will propose, it also decides on the maximum number of transactions it will include. That's my understanding. Is that correct?

Since 1 tx can only create or update 1 game at a time, it follows that there are a limited number of games updated in each block.

b9lab-content/3-main-concepts/10-multistore-keepers.md
xavierlepretre: Here it is a bit abstract, a codey way of introducing the numbers.

b9lab-content/3-main-concepts/10-multistore-keepers.md
xavierlepretre: Yep. Added.

b9lab-content/3-main-concepts/10-multistore-keepers.md
xavierlepretre: Updated.

b9lab-content/3-main-concepts/10-multistore-keepers.md
```diff
@@ -52,17 +52,17 @@ Keepers also implement an iterator method:
```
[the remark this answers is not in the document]
xavierlepretre: Fixed.

b9lab-content/3-main-concepts/10-multistore-keepers.md
nooomski: @xavierlepretre  The max block size (in bytes) is an on-chain consensus parameter that can be changed through governance. Validators themselves have no choice here and cannot increase the size. So I would reword this as follows:

```suggestion
Fortunately, the timeout duration is fixed and the same for all games. This means that those `n` games that all expire in a given block have all been created or updated at roughly the same time. Or in effect, at roughly the same block height `h` give or take a margin of error `h-1` and `h+1`. These created and updated games are in limited number because, as per one of the chain consensus parameters, every block has a maximum size and a limited amount of transactions it can include. So if by any chance, all games in blocks `h-1`, `h` and `h+1` expire now, then the `EndBlock` function would have to expire 3 times as many games as a block can handle. This is the worst case scenario, and still sounds manageable.
```

nooomski (commented):
Clarified block size limitations

PR #867, opened 2021-11-18
Final check of PR 811

CitMC:
Final check of PR 811 to handover for a final check by Cosmos before publication.


CitMC:
Merged for final review by the Cosmos team before publication

PR #869, opened 2021-11-18
Final check of PR 801

CitMC:
Final check of PR 801 to handover for a final check by Cosmos before publication.


CitMC:
Merge for final review by the Cosmos team before publication.

PR #893, opened 2021-11-29
testnet: update seeds ports

helder-moreira:
Thes PR updates seed ports with the correct ones after reseting the testnet for Hackatom.


helder-moreira:
> @helder-moreira Is this testnet info relevant outside of the Hackathon?

This info is for the gaia testnet we have been maintaining for a while now. It was used in Hackatom but was not created for it. Now I see that docs changed a lot, so it might not be relevant anymore. Feel free to close this if thats the case.

xavierlepretre:
@coldice How could we keep this info, that I think is valuable?

PR #894, opened 2021-11-29
docs: B9lab content updates release candidate 1

coldice:
This adds the Cosmos Academy content to the platform and implements all b9lab content and platform updates. All content has been pre-screened, internal language check is complete.

Preview is up on [https://deploy-preview-894--sdk-tutorials.netlify.app/](https://deploy-preview-894--sdk-tutorials.netlify.app/)


nooomski (approved):
Looks good to me, few easy things that need a fix before we launch:

- Can we remove the Atlas section in cosmos-ecosystem.html? This product isn't really being used right now and is missing a lot of modules. We can put it back at a later time.
- Main Concepts [Start Here] links to Accounts instead of Blockchain app architecture.
- Font & link colors are kinda botched in some of the image hotspot texts.
- We need to change the wording in Bridges. Gravity Bridge is not a cosmos product, so we should not call it the Cosmos Gravity Bridge but just "the Gravity Bridge". We also need to make it clear this is still in development and not yet launched before people get too excited.
- My Own Cosmos Chain [Start Here] links to Can Play Query instead of Starport.
- Not sure if it should happen before launch, but we discussed it earlier: can we make sure that we note which Starport version should be used and that this tutorials is designed for that version?

Some things that we should fix after launch:
- Broken link formatting in Terra link in cosmos-ecosystem.html
- Add synopsis to app architecture, accounts & transactions
- Queries is still using the old info highlight as a synopsis & in general could probably be a little expanded on in the future: what type of queries exist (grpc, rest, cli) and a sort of general overview for those who don't want to read the full query lifecycle. We can also just consider copying that doc tbh.
- Some of the wording in the synopsises needs a review, very often it's just "In this section you'll learn about [title of page]" without giving more introductory context. Would suggest something like "It is time to address how data storage is handled in the Cosmos SDK using multistore and keeper." instead of "It is time to address multistore and keeper." for example.
- We should probably upgrade Migrations to contain something about cosmovisor.
- Bridges still using the old info highlight synopsis.
- Intro to "running a chain" has "Let’s take it step-by-step with simppp.", should be simApp
- Running a Node could use another editor's review.
- The titles of the articles in My Own Cosmos Chain could probably use a revisit. "Creating a game message" instead of "create a game message", "Making a Move" v.s. "The Play Game Elements", "Managing Your Game List" instead of "A Game FIFO", etc.
- The articles in My Own Cosmos Chain would also need another editor's check imo. Some of the language is still a bit off, with some sentences still being either overly complex, too short, or not fitting the voicing as discussed before. Would love for this to be a **priority** after launch. Some synopsis also need more work, what are we building on this article?
- Some of the "next up" in the exercises refer you to different sections if you want to skip some parts. Can we confirm that going straight to those sections will actually work? Has anyone tested this? Would be great if we list these connections and get someone to test these build orders to confirm nothing breaks.
- The big yellow warning sign in game-wager.html is probably a bit too much, wdyt?
- Would love to include the typescript relayer setup in wager-denom.html
- In general the exercise could use the free front-end that comes with Starport a bit more (localhost:8080) to send tokens or use the API (localhost:1317) to test out some of the queries as you build your chain & implement messages. This testing out your work is essential to making the user feel satisfied imo.
- We should probably specify that the CosmWasm tutorial is not related to the checkers game.

PR #923, opened 2021-12-12
fix conditional expression

whois1:


scavenge/tutorial/07-keeper.md
```diff
@@ -177,7 +177,7 @@ func (k msgServer) RevealSolution(goCtx context.Context, msg *types.MsgRevealSol
   // check that the scavenger property contains a valid address
 	_, err := sdk.AccAddressFromBech32(scavenge.Scavenger)
   // return an error if a scavenge has already been solved
-	if err == nil {
+	if err != nil {
```
xavierlepretre, line 180: Disclaimer: I did not write this code.
However, to me the original looks correct. If the scavenge has already been solved, then it contains the `msg.Creator` as per line 184 below. Therefore, if it has already been solved, it need not be solved again.
And if it has been solved, then the `.Scavenger` address is well formed and does **not** return an error when parsed.

PR #936, opened 2021-12-16
Added wording at the end of migrations.

xavierlepretre:


xavierlepretre:
@coldice Is it possible to have a visual break on the left-hand outline so that CosmJs is separated from migrations?

coldice:
> @coldice Is it possible to have a visual break on the left-hand outline so that CosmJs is separated from migrations?

We have added a new feature for this, you can add
`divider: true` to the frontmatter header of the content page you want the divider to appear at.

CitMC:
@coldice can be merged

coldice (approved):

coldice (approved):

PR #960, opened 2022-01-10
Remove double text in multistore-keepers.md

nooomski:
Some double text that should be removed.


xavierlepretre (approved):

ademarco:
I was about to open the exact same PR! :+1:

PR #1041, opened 2022-04-26
Gb new update

gioelebonifetto:
* Updates the module langinpage template to handle frontmatter
* implements panel list component
* updates the google tag manager


xavierlepretre:
Minor remark. The line on the left ends abruptly and it is unclear whether it ended because that was the end of the point, or because the display has some issue.
Can it be made to end less abruptly? For instance, it could end with an inverted T.
Screenshot 2022-04-26 at 19 46 41

xavierlepretre (approved):

gioelebonifetto:
@xavierlepretre Thank you for the comment, I agree with you now it looks cleaner
Screenshot 2022-04-26 at 19 16 36
Screenshot 2022-04-26 at 19 16 46

coldice (approved):

PR #1042, opened 2022-04-27
IBC connections draft PR

charleenfei:
This PR splits up the TAO layer into the three relevant components: Connections, Channels, and Clients.

It goes into a deep dive on the Connections component. Formatting is not final, but content is there.

### Change scope

- [ ] Small content fixes 
- [x] Addition of new content
- [ ] Platform fixes
- [ ] Other


academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+// the connection handshake.
+type Version struct {
+	// unique version identifier
+	Identifier string `protobuf:"bytes,1,opt,name=identifier,proto3" json:"identifier,omitempty"`
+	// list of features compatible with the specified identifier
+	Features []string `protobuf:"bytes,2,rep,name=features,proto3" json:"features,omitempty"`
+}
+```
+
+Protocol versioning is important to establish, as different protocol versions may not be compatible due to ie: proofs being stored on a different path. There are three types of protocol version negotiation:
```
tmsdkeys, line 41: for example instead of ie

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+```
+
+In this definition, `connection-id` is used to as a key to map and retrieve connections associated with a certain client from the store. 
+
+`prefix` is being used by the clients to construct merkle prefix paths which are then used to verify proofs.
+
+
+**Connection Handshakes and States**
+
+Establishing an IBC connection between example chain A and chain B requires four handshakes. A high level overview of a successful four-way handshake:
```
tmsdkeys, line 105: Should we not call it a 4-step handshake? So not 4 handshakes

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+    return nil, sdkerrors.Wrap(err, "connection handshake open init failed")
+  }
+
+  return &connectiontypes.MsgConnectionOpenInitResponse{}, nil
+}
+```
+
+<HighlightBox type="info">
+
+In IBC, blockchains do not directly pass messages to each other over the network. To communicate, blockchains commit the state to a precisely defined path reserved for a specific message type and a specific counterparty. In the case of establishing a connection,  Relayers monitor for updates on these paths and relay messages by submitting the data stored under the path along with proof of that data to the counterparty chain.
```
tmsdkeys, line 180: should this be in here? looks a different tone than the rest

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+  // Check that ChainA stored the correct ConsensusState of chainB at the given consensusHeight
+  if err := k.VerifyClientConsensusState(
+    ctx, connection, proofHeight, consensusHeight, proofConsensus, expectedConsensusState,
+  ); err != nil {
+    return "", err
+  }
+```
+
+so both will verify the `ConnectionState`, the `ClientState` and the `ConsensusState` of the other chain.
+
```
tmsdkeys, line 283: Add that the connection state went from TRY to OPEN on chain A

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+</HighlightBox>
+
+**OpenTry**
+
+This `OpenInit` will be followed by an `OpenTry` response. in which chain B will verify the identity of chain A according to information that chain B has about chain A in its light client (the algorithm and the last snapshot of the consensus state containing the root hash of the latest height as well as the next validator set), as well as respond to some of the information about its own identity in the `OpenInit` announcement from chain A. 
+
+![OpenTry](/academy/ibc/images/open_try.png)
+
+The purpose of this handshake is not only for chain B to verify that the counterparty chain A is indeed the expected counterparty identity, but also to verify that the counterparty has the accurate information about chain B's identity. The relayer will also submit two `UpdateClient`s with chain A and chain B as source chains before this handshake. These `UpdateClient` will update both chain A and chain B light clients, in order to make sure that the state verifications in this step are successful.
```
tmsdkeys, line 192: Add term 'double verification' + "The purpose of this step of the handshake..." instead of "The purpose of this handshake'

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+  // counterparty chain associated with this connection.
+  Counterparty counterparty = 4 [(gogoproto.nullable) = false];
+  // delay period that must pass before a consensus state can be used for
+  // packet-verification NOTE: delay period logic is only implemented by some
+  // clients.
+  uint64 delay_period = 5 [(gogoproto.moretags) = "yaml:\"delay_period\""];
+}
+```
+
+`ConnectionEnd` defines a stateful object on a chain connected to another separate one. Note that there must only be 2 defined `ConnectionEnd`s to establish a connection between two chains. So the connections are mapped and stored as `ConnectionEnd` on the chain.
```
tmsdkeys, line 74: `ConnectionEnd` defines a stateful object on a chain connected to its counterpart on the counterparty chain. Note that there must only be 2 defined `ConnectionEnd`s to establish a connection between two chains. So the connections are mapped and stored as `ConnectionEnd` on the respective chains.

academy/ibc/ibc-tao-dev.md
charleenfei: will reference in another section

academy/ibc/ibc-tao-dev.md
```diff
@@ -0,0 +1,314 @@
...
+  EmitConnectionOpenInitEvent(ctx, connectionID, clientID, counterparty)
+
+  return connectionID, nil
+}
+
+```
+
+This function creates an unique `connectionID`. You can see that it adds the connection to a list, which represents the connections associated with a specific client.
+
+Also you can see how it creates a new `ConnectionEnd`:
```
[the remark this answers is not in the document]
charleenfei: revised

coldice (approved):

coldice:
ACK @CitMC

PR #1054, opened 2022-05-07
updates info and warning highlight boxes

gioelebonifetto:
**Instructions:** Please fill out the following items in brackets `[ ]`, check all items in the change scope list below and remove this line before submmitting your Pull Request.

Documentation content update for [Module X, Section Y]

This PR [... (elaborate here)]

[(Optional: Further notes for reviewers)]

### Change scope

- [ ] Small content fixes 
- [ ] Addition of new content
- [ ] Platform fixes
- [ ] Other


.vuepress/components/HighlightBox.vue
```diff
@@ -84,13 +84,8 @@
...
 
         &.warn, &.warning {
-            background: var(--color-warning);
-            color: black
-            fill: black
-
-            .title {
-                color: black
-            }
+            background: linear-gradient(78.06deg, #00B067 0.2%, #FFD303 97.95%);
```
xavierlepretre, line 87: Maybe this is splitting hair:

1. A value like `78.06deg` makes it look like it was created approximatively. Does it matter if it was `78deg`?
2. Line 77 we have `98.88%` as the last value, and here we have `97.95%`. Were they intended to be equal but you set them with a slider instead of precise values?
3. Not sure if it makes sense in this case, but it could be `const gradientAngle = "78 deg";` and reuse it. Same for other repeated values.

coldice (approved):

PR #1125, opened 2022-05-29
Update Ethereum gas price unit

jrstrunk:
Documentation content update for the Public and Managed Blockchains page in the Week 0 section of the Interchain Developer Academy.

This PR updates the mentioned unit of Ethereum gas prices.

Reading through the documentation I noticed it said "any **Gas price** in terms of **Ether**, a value which Ethereum requires to convert **Ether to Gas**". Using Ether twice in this way sounded a little weird to me and normally Ethereum gas prices are specified in Gwei, so I thought this may have just been a typo. Although since Gwei is just Ether with a shifted decimal place, I could see an argument to keep it how it was as maybe the original meant to keep it simple for new users.

### Change scope

- [x] Small language/grammar fixes
- [ ] Small content fixes 
- [ ] Addition of new content
- [ ] Platform fixes
- [ ] Other

### DIFF
--- a/academy/0.0-B9lab-Blockchains/3_managed.md
+++ b/academy/0.0-B9lab-Blockchains/3_managed.md
@@ -56,7 +56,7 @@ The most important difference compared to Bitcoin is the implementation of distr
 
 The code execution platform is [Turing-complete](https://en.wikipedia.org/wiki/Turing_completeness). As part of Turing-completeness, frameworks must overcome the [halting problem](https://www.scientificamerican.com/article/why-is-turings-halting-pr/), which is especially difficult in distributed, hierarchy-free computing platforms. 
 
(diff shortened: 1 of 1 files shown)


xavierlepretre:
Yes, those in the field say it is in Gwei, but that's an accident of its current price. If the gas price increases from around 10^-9 Ether to around 10^-6 Ether, we would switch to Terawei. Conversely if the gas price fell to around 10^-12, then everybody would start talking in Megawei, not 0.001 Gwei.
And the wei and its Gwei and such are probably less well-known than the Ether term for beginners.
Pedantic alert: The gas price's unit really is Ether/gas (or Gwei/gas).

jrstrunk:
Makes sense, just wanted to have someone look at it again to make sure it wasn't a typo. I'll go ahead and close this PR then since using Ether as the unit for gas is more beginner friendly than using Gwei as the unit for gas.

PR #1127, opened 2022-05-30
docs: fix GUI run instructions

SpiralOutDotEu:
Documentation content update for [academy, 4-my-own-chain]

This PR fixes the `npm run` command sample and corrects the `localhost` port

### Change scope

- [ ] Small content fixes 
- [ ] Addition of new content
- [x] Sample code/command updates
- [ ] Platform fixes
- [ ] Other

### DIFF
--- a/academy/4-my-own-chain/ignitecli.md
+++ b/academy/4-my-own-chain/ignitecli.md
@@ -266,10 +266,10 @@ Boot up the frontend created by Ignite CLI by using the commands provided in the
 ```sh
 $ cd vue
 $ npm install
-$ npm run serve
+$ npm run dev
 ```
 
-Navigate to [localhost:8080](http://localhost:8080/). On the client-side, no wallets have been created or imported yet. Load Alice's wallet in the GUI to have some tokens. You will need to use the mnemonic for Alice, which you can find in the output of the `ignite chain serve` command. Copy and paste it to _import a wallet_.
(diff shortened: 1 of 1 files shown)


coldice:
Hey @xavierlepretre, can you please verify this fix? npm run command for the vue app, should call `dev` instead of `serve` and runs on `3000` not `8080` - correct?

SpiralOutDotEu:
Hi @xavierlepretre, `package.json` is not as you describe. 
I think it changed after moving from starport to ignite-cli, but now it uses `Vite`

In `package.json` you will find:
```
"scripts": {
    "dev": "vite",
    "build": "vite build",
    "serve": "vite preview"
  },
```
According to the above, the correct command should be:
```
$ npm install
$ npm run dev
```
**or**
```
$ npm install
$ npm run build
$ npm run serve
```

I've chosen the first as it is faster and more clear that you are working on the development environment.

xavierlepretre:
Ah ok then this change refers to Ignite's new way of doing. I suppose you have Ignite 0.21.3 or something close to it. The documentation is still based off the 0.17.3 version.

So you are already in effect with the _newer version_ I mentioned at the end of my comment.

We will update the versions of the code project and the tutorials in our next round of updates.

SpiralOutDotEu:
Indeed, `0.21.3` is my version and everything in this doc works as expected besides these commands.
I saw also some people in discord struggling with this and that's why I started this PR.

PR #1164, opened 2022-07-10
feat: add group tutorial

julienrbrt:
This PR adds the `group` module tutorial.

It as well moves the `authz` tutorial under `/tutorials` for consistency.
It formats the Markdown of the existing tutorials.

### Change scope

The changes in this Pull Request include (please tick all that apply):

- [x] Small language/grammar fixes
- [x] Small content fixes 
- [x] Addition of new content
- [ ] Sample code/command updates
- [x] Platform fixes
- [ ] Other

PR #1167, opened 2022-07-13
feat: enforce markdownlint in the repository

julienrbrt:
Follow-up of  https://github.com/cosmos/sdk-tutorials/pull/1164#discussion_r919996384

The `.markdownlint.json` configuration has been introduced, but is not yet enforced.
This PR aims to enforce the rule in all the files after having discussed the rules present in `.markdownlint.json` (so not all files have been touched yet).

The list of available rules is available here: https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md

### Change scope

- [ ] Small content fixes 
- [ ] Addition of new content
- [ ] Platform fixes
- [ ] Other

### DIFF
--- a/.markdownlint.json
+++ b/.markdownlint.json
@@ -3,10 +3,12 @@
     "MD004": {"style": "asterisk"},
     "MD007": { "indent": 4 },
     "MD013": false,
+    "MD014": false,
     "MD024": { "siblings_only": true },
     "MD025": false,
     "MD033": false,
     "MD034": false,
+    "MD049": {"style": "underscore"},
     "no-hard-tabs": false,
     "whitespace": false
 }
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,3 @@
+{
+  "recommendations": ["DavidAnson.vscode-markdownlint"]
+}
--- a/course-ida/discord-info.md
+++ b/course-ida/discord-info.md
@@ -13,13 +13,14 @@ There are six channels - each dedicated to a specific topic - where you can post
 ## How to join the Academy Discord
 
 Follow these steps to join the Academy Discord:
+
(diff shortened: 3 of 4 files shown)


.markdownlint.json
```diff
@@ -3,10 +3,12 @@
     "MD004": {"style": "asterisk"},
     "MD007": { "indent": 4 },
     "MD013": false,
+    "MD014": false,
```
xavierlepretre, line 6: @coldice This one is about [`$` in code blocks](MD014). Not sure what `false` means here.
@julienrbrt The rationale mentioned is that it is easier to copy without the `$` but our copy button removes them.
I am in favour of having the `$`.

.markdownlint.json
```diff
@@ -3,10 +3,12 @@
...
     "MD024": { "siblings_only": true },
     "MD025": false,
     "MD033": false,
     "MD034": false,
+    "MD049": {"style": "underscore"},
```
xavierlepretre, line 11: This one is for emphasis (italics), we have written them with `_` throughout so we ought to make it official.

.vscode/extensions.json
```diff
@@ -0,0 +1,3 @@
+{
+  "recommendations": ["DavidAnson.vscode-markdownlint"]
```
xavierlepretre, line 2: @coldice Like this everyone is prompted to install it.

.markdownlint.json
julienrbrt: The `false` disable the rule :)

xavierlepretre: So there is no way to enforce the presence of `$`? Only a way to enforce (or not) the absence of `$`?

julienrbrt: Nope this rule can only be enabled of disabled

xavierlepretre (approved):

PR #1199, opened 2022-09-02
Add checkers extension

anilder89:
Documentation content update for [Module IBC dev, Checkers Extension]
### Change scope

* [ ] Small language/grammar fixes
* [ ] Small content fixes 
* [X] Addition of new content
* [ ] Sample code/command updates
* [ ] Platform fixes
* [ ] Other


academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,342 @@
...
+  repeated PlayerInfo player = 1; 
+  string GameID = 2; 
+}
+```
+
+This way, you keep also track of the `GameID` and as you can guess, the `leaderboard` chain will maintain different leaderboards with different `GameID`s for different chains.
+
+So you can use the same logic as in the `checkers` chain to send and receive packets. The difference is that for the `leaderboard` chain, the only way to get into one of it's leaderboard is by sending score packets from a game chain.(In `checkers` chain, you will get into the leaderboard structure if you win enough games.)
+
+Anyway we will keep the sending part untouched because we are interested in receiving packets. In the `x/leaderboard/keeper/ibc_top_rank.go` you can find almost the same code as in the `x/checkers/keeper/score.go`:
```
xavierlepretre, line 188: I would still let the Leaderboard chain send score packets, because it could be sending them to other leaderboard chains.

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,342 @@
...
+...
+```
+
+## Adjust `leaderboard`
+
+In the `leaderboard` chain, you need first to do the same steps as in the [Migration](add_link) section for the `checkers` chain and include a leaderboard but you will need a slightly different leaderboard structure for the `leaderboard` chain:
+
+```
+message Board {
+  repeated PlayerInfo player = 1;
```
xavierlepretre, line 179: In Checkers, `PlayerInfo` refers to something other than how it is defined here. Perhaps we want to prevent confusion and rename it there to `WinnerInfo` or `PlayerInLeaderboard`.

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,342 @@
...
+
+```bash
+$ docker exec -it leaderboard bash
+```
+
+and check the board:
+
+```bash
+$ leaderboardd q leaderboard list-board
+```
```
xavierlepretre, line 342: Looks incomplete but I have not run the project to know what it will output.

xavierlepretre (changes requested):
I wonder how much time we have to modify the code of the project too.

I have an idea for the next iteration of this example is to have checkers create a new module for leaderboard. It could be reusable to a better extent.

academy/ibc-dev/ibcapp_checkers.md
anilder89: Yeah that is a use case I did not think of, thanks.

academy/ibc-dev/ibcapp_checkers.md
anilder89: You mean it would be better to have the output included?

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,342 @@
...
+
+  if !found {
+    panic("Leaderboard not found")
+  }
+
+  // Construct the packet
+  var packet types.ScorePacketData
+
+  found_in_leaderboard := false
+  for i := range leaderboard.Winners {
```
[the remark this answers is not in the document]
anilder89: Good thought. I wonder if it is better to keep the list(and to sort it after each update) and just use a binary search or if it is better to implement a binary search tree structure...

tmsdkeys (commented):
Hi @anilder89 thanks for your work on this and apologies for not being able to review sooner.

Before going more in detail, I want to first get some more clarity on what we are doing high-level with the checkers project.

This is how I saw the general narrative outlining the iterations:

1. The initial checkers blockchain (pre-migration), `x/checkers` was not IBC-enabled but does allow for wagers with IBC denoms
2. To showcase migrations, a leaderboard data structure was added to the `x/checkers` module
However, I think this is not the preferred way when thinking about moving forward as I think the leaderboard functionality could be in its own module, staying aligned with the vision of the SDK as being modular and separating distinct application logic that could be re-used, in this case a `x/leaderboard` could be used by many game modules/chains

  I believe @xavierlepretre had this in mind also:
  > I have an idea for the next iteration of this example is to have checkers create a new module for leaderboard. It could be reusable to a better extent.

3. (3a) A next iteration could focus on single-chain functionality with both `x/checkers` and `x/leaderboard` modules on the same chain. Starting from an earlier commit pre-migration
4. (3b)An iteration taking a different approach would be the IBC enabled `x/checkers` module that outsources the leaderboard functionality to a new chain (and as a result, does not include an own `x/leaderboard` module nor leaderboard data structure in `x/checkers`). The idea being here that the leaderboard chain is an appchain specifically for aggregating a general leaderboard for gamers of all types of different game (chains), adding weights to different games according to their difficulty, length, ...
5. (This is an optional iteration on the IBC path when Interchain Queries are finished, we could even play from different chains with ICA and ICQ)

Right now, the `x/checkers` module still contains a local leaderboard structure and the leaderboard chain also stores these local leaderboards (as you separate base on `GameId` or `GameTypeId`), not an aggregated leaderboard of all possible game chains... I'm a bit confused as to why we'd do that

academy/ibc-dev/ibcapp_checkers.md
tmsdkeys: My thinking behind only letting checkers send score packets to leaderboard and restricting the leaderboard chain to receiving them, is that we can showcase asymmetric IBC packet flows (Leaderboard throws an error if trying to send or acknowledge a `sendScore` packet

tmsdkeys:
Additionally, I think it makes sense to make the hands-on a bit more modular. What I mean is that if someone wanted to only do the checkers IBC extension, this tutorial could focus on IBC functionality by doing the following:

- Build the checkers blockchain from the commit pre-migration (so without leaderboard data structure added) or includes a script to run through it (wanted to look into how feasible that is) if we want to give people insight that may not have followed the checkers exercise (we could provide them with a checkers version built with the --ibc flag from the start)
- Explain how to use checkers from the CLI
- Include those scripts that simulates a game
- Explains how to build the leaderboard chain
- Then focuses on implementing the channel and packet callbacks etc. when using IBC

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,341 @@
...
+}
+```
+
+As you can see, instead of sending an arbitrary score for an arbitrary player, the system first verifies that the player appears on the leaderboard. Use the `ChannelID` as the unique `GameID`. Now you need to modify the `MsgSendScore` in the `proto/tx.proto` and make it to:
+
+```protobuf
+message MsgSendScore {
+    string sender = 1;
+    string port = 2;
+    string channelID = 3;
```
tmsdkeys, line 97: The `channelID` here would be the source channel one would specify to send the IBC packet right, so it would not necessarily be unique on the leaderboard chain, correct?

You could only use counterparty `channelID` as a unique game identifier, but even then what happens when a relayer accidentally creates another channel?

murataniloener:
@tmsdkeys @xavierlepretre 

After doing some experimenting I think this is how we should process:
https://github.com/b9lab/cosmos-ibc-docker/tree/ao-modular/modular/b9-checkers-academy-draft

It extends the checkers game with a leaderboard module, which stores the player information and maintains a leaderboard.
- playerInfo is a part of the leaderboard module because this information is only useful for the leaderboard.
- I do not like the idea of sorting and clipping on the chain, because it costs a lot and all playerInfo is maintained anyway, so a task that can be solved better on the client side. Anyway, a leaderboard module should maintain a leaderboard, so there is a separate transaction for a leaderboard update. 

The leaderboard module is like a plug-in, the checkers game will use it if a game is over and send the information about a win/loss/draw to the leaderboard module. 

At this point, I think instead of making both modules more separate and letting checkers game send an IBC packet to the leaderboard, it is better to let it like it is now and extend the leaderboard module to make it able to send/receive playerInfo:
- I think the idea of sending each win/loss via a IBC packet is not ideal, instead it makes more sense to send a playerInfo packet.
- there is no need for the checkers game to maintain a list of playerInfo, it only makes sense if you consider a leaderboard. 
- no other use case of IBC for the checkers game besides the leaderboard. So no need for the checkers module to be IBC enabled. 

I think it is better, if we see the leaderboard module as an extension needed to connect to the global leaderboard. We can do something symmetric,  like the leaderboard module maintains a local leaderboard and can send playerInfo to a global leaderboard or we can do a more asymmetric approach and remove the board structure from the leaderboard module, only let it store the playerInfo(or perhaps rename the module into playerstats or something like that) and create another module/chain for the global leaderboard, which will have a leaderboard structure, waiting for playerInfo to be sent.

What do you think?

xavierlepretre:
1. I agree that the leaderboard module should be a separate module.
2. About taking `PlayerInfo` out of the checkers module: I understand that this is something that could be taken out of checkers and kept in a Web2 server, although you lose the certainty about the values. On the other hand, the benefit I could see of keeping it in is that it can be said to help players compare each other 1 on 1, and perhaps to create leagues. Also, I was hoping to split the migration tutorial into 2 parts, in order to reduce the size of the learning and to drive home the difference between module consensus version and app version:
    - part 1: consensus version goes from 2 to 3, and `PlayerInfo` is introduced and computed on upgrade. Some hand-waving saying if you want to upgrade only this, then it is possible. Tests.
    - part 2: consensus version goes from 3 to 4, and the new leaderboard module is introduced and computed on upgrade. Some hand-waving saying that actually the upgrade proposal that passed includes both upgrades at the same time, and call the app upgrade from v1 to v2, so we have to prepare the code for that. Tests.
3. About sorting and clipping on chain, I actually had an idea for optimisation and further learning. The leaderboard would be read and loaded in BeginBlock, then each transaction updates or adds (same code as now), then in EndBlock, we sort, clip and store. This leaderboard handling will also have to happen after the checkers' own EndBlock, which may forfeit games, i.e. there is a new winner. With this, you demonstrate the importance of the call order of EndBlocks, and there are fewer read/write and a single sorting per block. That would present another use of these "strange" parts. Of course, I need to confirm that a custom object can stay in memory (the context?) from BeginBlock to EndBlock.
4. How big a concern are IBC costs if the leaderboard is not even on the checkers chain?

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,525 @@
...
+For the code above to function, you need to define `TimeLayout` in the `x/leaderboard/types/keys.go`. Add the following piece of code at the end of the file:
+
+```golang
+const (
+    TimeLayout              = "2006-01-02 15:04:05.999999999 +0000 UTC"
+    LeaderboardWinnerLength = uint64(100)
+)
+```
+
+Now it is time to allow the checkers module access to the leaderboard module. Therefore, look for the `app.CheckersKeeper` in `app/app.go` and modify it in order to include `app.LeaderboardKeeper`:
```
tmsdkeys, line 156: This is a nice consequence of the approach you went for, allowing to review passing capabilities to another module :)

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,525 @@
...
+    ctx := sdk.UnwrapSDKContext(goCtx)
+
+    playerInfoList := k.GetAllPlayerInfo(ctx)
+    k.updateBoard(ctx, playerInfoList)
+
+    return &types.MsgUpdateBoardResponse{}, nil
+}
+```
+
+That is it! Now the checkers blockchain can keep track of the player information, and create or update the leaderboard based on the player information if requested via the CLI.
```
tmsdkeys, line 355: I'd make a split here and start a new section/file to discuss IBC specific parts vs the prep that is more checkers/SDK focused

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,525 @@
+# Extend the Checkers Game With a Leaderboard
+
+Go to your checkers folder and make sure that you checked out of the [v1-cosmjs-elements](https://github.com/cosmos/b9-checkers-academy-draft/tree/v1-cosmjs-elements) tag.
```
tmsdkeys, line 3: This is a commit from the earlier checkers project with ignite cli v0.17, which gives a notification to migrate from Ignite CLI when attempting next scaffolding step with Ignite CLI v0.22

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,525 @@
...
+
+    rules "github.com/b9lab/checkers/x/checkers/rules"
+    "github.com/b9lab/checkers/x/checkers/types"
+
+    sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+func getWinnerAndLoserAddresses(storedGame *types.StoredGame) (winnerAddress sdk.AccAddress, loserAddress sdk.AccAddress) {
+    if storedGame.Winner == rules.PieceStrings[rules.NO_PLAYER] {
+        panic(types.ErrThereIsNoWinner.Error())
```
tmsdkeys, line 61: Given comment above, I started from another commit so this might explain, but I had to register this error `ErrThereIsNoWinner`

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,525 @@
...
+        panic(err.Error())
+    }
+    if storedGame.Winner == rules.PieceStrings[rules.RED_PLAYER] {
+        winnerAddress = redAddress
+        loserAddress = blackAddress
+    } else if storedGame.Winner == rules.PieceStrings[rules.BLACK_PLAYER] {
+        winnerAddress = blackAddress
+        loserAddress = redAddress
+    } else {
+        panic(fmt.Sprintf(types.ErrWinnerNotParseable.Error(), storedGame.Winner))
```
tmsdkeys, line 78: Given comment above, I started from another commit so this might explain, but I had to register this error `ErrWinnerNotParseable`

academy/ibc-dev/ibcapp_checkers.md
xavierlepretre: Replace with branch `cosmjs-elements` which is the current version.

murataniloener: Indeed I used the cosmjs-elements branch for the test I did at https://github.com/b9lab/cosmos-ibc-docker/tree/ao-modular/modular, but I wanted to have a tag like the one you have for v1. So I have been keeping in mind to replace this link as soon as you have a v2 tag, which I assumed would be included for the next cohort.

academy/ibc-dev/ibcapp_checkers.md
murataniloener: Note for me: scan for panics in the code and replace them

academy/ibc-dev/ibcapp_checkers.md
xavierlepretre: My branches are still subject to change a little, which is why I'd rather only create the v2 tags when working on v3 😬.
FYI, if you look at the commits, you see that each branch has 2 commits. It was a long and arduous process with a lot of squashing and reordering, but it makes for a cleaner "history". 
I want to keep it this way, so the content of the branch may change although the commits' dates and positions don't.

academy/ibc-dev/ibcapp_checkers.md
anilder89: Good catch, was missing in the correct commit too.

academy/ibc-dev/ibcapp_checkers.md
```diff
@@ -0,0 +1,533 @@
...
+
+You need to do some more work to update the board with that information.
+
+There are two places at which you can call for an update on the board structure, in `OnRecvCandidatePacket` so each player sending his/her information will pay the fee for sorting and clipping the leaderboard, or you can again create a separate transaction for anyone to sort and clip the leaderboard on the leaderboard chain like you did for the checkers chain.
+
+Take this as an opportunity to practice and implement it in `OnRecvCandidatePacket`.
+
+## Test it
+
+You can find the sample implementation of the checkers chain extension and the leaderboard chain on a [repository we provide](https://github.com/b9lab/cosmos-ibc-docker/tree/ao-modular/modular). There you will also find a Docker network and the relayer settings for an easy test. In addition, the repository includes a script to create and run games. Follow the steps described in the repository to run a few tests and to see it in action.
```
tmsdkeys, line 533: This content is now duplicate right?

academy/ibc-dev/ibcapp_leaderboard.md
```diff
@@ -0,0 +1,182 @@
+## Create a Leaderboard chain
+
+After the extention of the checkers chain with a leaderboard module, the checkers game is able to keep track on the player stats and it can maintain(on request) a sorted leaderboard. In addition it can send player stats via IBC to another chain. 
+You will now create a leaderboard chain which can receive those packages. Determine another folder for your leaderboard chain, and scaffold a chain via Ignite:
```
tmsdkeys, line 4:
```suggestion
You will now create a leaderboard chain which can receive the `Candidate` packets to store a global leaderboard. Determine another folder for your leaderboard chain, and scaffold a chain via Ignite CLI:
```

academy/ibc-dev/ibcapp_leaderboard.md
```diff
@@ -0,0 +1,182 @@
...
+$ ignite scaffold map playerInfo wonCount:uint lostCount:uint dateUpdated:string --module leaderboard --no-message
+```
+
+and of course a board structure:
+
+```bash
+$ ignite scaffold single board PlayerInfo:PlayerInfo --module leaderboard --no-message
+```
+
+In addition, you want to receive candidate packages:
```
tmsdkeys, line 28:
```suggestion
In addition, you want to receive candidate packets:
```

academy/ibc-dev/ibcapp_leaderboard.md
```diff
@@ -0,0 +1,182 @@
...
+
+```bash
+ignite scaffold packet candidate PlayerInfo:PlayerInfo --module leaderboard --no-message
+```
+
+this time you use the `--no-message` flag because this chain is not going to send any player information to another chain.
+
+Implement the logic for packet receivment in the `x/leaderboard/keeper/candidate.go`:
```
tmsdkeys, line 36:
```suggestion
Implement the logic for receiving packets in the `x/leaderboard/keeper/candidate.go`:
```

academy/ibc-dev/ibcapp_leaderboard.md
```diff
@@ -0,0 +1,182 @@
...
+```golang
+...
+// OnRecvCandidatePacket processes packet reception
+func (k Keeper) OnRecvCandidatePacket(ctx sdk.Context, packet channeltypes.Packet, data types.CandidatePacketData) (packetAck types.CandidatePacketAck, err error) {
+    // validate packet data upon receiving
+    if err := data.ValidateBasic(); err != nil {
+        return packetAck, err
+    }
+
+    // TODO: packet reception logic
```
tmsdkeys, line 47: Think this can be dropped

academy/ibc-dev/ibcapp_leaderboard.md
```diff
@@ -0,0 +1,182 @@
...
+package types
+
+import (
+    "errors"
+)
+
+// ValidateBasic is used for validating the packet
+func (p CandidatePacketData) ValidateBasic() error {
+
+    // TODO: Validate the packet data
```
tmsdkeys, line 81: same here

tmsdkeys:
Thanks for the work to add the logic for updating the board in the `OnRecvPacket` callback @anilder89 

Left a few additional comments.

Since this PR has been open for some time now, and still contains the old portal structure, would it not be best to close this PR and open a new one with the latest version + the two readme files with the new content? cc @CitMC

AndrewB9lab (approved):
Language review completed.

anilder89:
> Thanks for the work to add the logic for updating the board in the `OnRecvPacket` callback @anilder89
> 
> Left a few additional comments.
> 
> Since this PR has been open for some time now, and still contains the old portal structure, would it not be best to close this PR and open a new one with the latest version + the two readme files with the new content? cc @CitMC I feel that would be cleaner

@CitMC I think the content is ready now. Can you please create another PR using the files from this one but checking out from master?

tmsdkeys:
@CitMC only two of the files: `ibcapp_checkers.md` and `ibcapp_leaderboard.md`
The others have already been merged into master prior to C2

CitMC:
All substantive changes were included in : https://github.com/cosmos/sdk-tutorials/pull/1290 PR can be closed.

coldice (approved):

PR #1204, opened 2022-09-08
feat: update landing page article and cherry-pick tutorials

julienrbrt:
The landing page is outdated:

- it displays old articles (one that even promotes Tendermint 0.35 (!) which has been since then discontinued)
- mention only one tutorial, IMHO we don't need the ignite card, as we mention it just below.

This PR updates the articles with hand-picked recent ones, add two tutorials instead of that big ignite card.

By the way, do you guys really like pug? 😅


xavierlepretre (approved):
When `npm rn serve`, the landing page lgtm

coldice:
There is an issue with the card height alignment on the Tutorials cards
image

PR #1231, opened 2022-09-29
Update IBC article in dev portal

nooomski:
Hey @coldice, running locally I get `Error: Cannot resolve theme: cosmos` but I only changed the link and image, so should be fine. Could you review? Thanks! :)


coldice:
Hey @nooomski I think the error was caused by a missing package (did you run `npm i`)? Anyways, fixed a small syntax issue, checked looks good 👍

PR #1464, opened 2023-10-06
docs: Custom cosmjs interfaces tutorial update

Zetazzz:
Documentation content update for [Section `Create Custom CosmJS Interfaces`]

This PR updates outdated `Create Custom CosmJS Interfaces` section, since the old tutorial of transpiling proto files(using protoc) is not beginner friendly. 
With the new tutorial, users can learn the process in [telescope](https://github.com/cosmology-tech/telescope) way, which is easier to install and transpile proto files.
Some links to [cosmjs-types](https://github.com/confio/cosmjs-types/) are also outdated(current version of cosmjs-types was also generated by telescope), so I updated those too.

[(Optional: Further notes for reviewers)]

### Change scope

* [ ] Small language/grammar fixes
* [ ] Small content fixes
* [x] Addition of new content
* [x] Sample code/command updates
* [ ] Platform fixes
* [ ] Other

### DIFF
--- a/tutorials/7-cosmjs/5-create-custom.md
+++ b/tutorials/7-cosmjs/5-create-custom.md
@@ -234,7 +234,7 @@ You can find the same structure in [`query.ts`](https://github.com/confio/cosmjs
 
 ### Proper saving
 
-Commit the extra `.proto` files as well as the compiled ones to your repository so you do not need to recreate them.So add an npm run target, to keep track of how this was done and easily reproduce it in the future when you update a Protobuf file:
+Commit the extra `.proto` files as well as the compiled ones to your repository so you do not need to recreate them. So add an npm run target, to keep track of how this was done and easily reproduce it in the future when you update a Protobuf file:
 
 ```json
 "scripts": {
--- a/tutorials/7-cosmjs/5-create-custom.md
+++ b/tutorials/7-cosmjs/5-create-custom.md
@@ -39,7 +39,7 @@ This exercise assumes that:
 
 1. The library you're creating is in 'myLib'.
 2. Your Protobuf definition files are in `./proto/`.
-2. You want to transpile them into TypeScript in `./src/codegen/`.
+3. You want to transpile them into TypeScript in `./src/codegen/`.
 
 Install `telescope` on your computer globally:
 
--- a/tutorials/7-cosmjs/5-create-custom.md
+++ b/tutorials/7-cosmjs/5-create-custom.md
@@ -31,119 +31,37 @@ This section assumes that you have a working Cosmos blockchain with its own modu
 
 ## Compiling the Protobuf objects and clients
 
-You can choose which library you use to compile your Protobuf objects into TypeScript or JavaScript. Reproducing [what Stargate](https://github.com/cosmos/cosmjs/blob/main/packages/stargate/CUSTOM_PROTOBUF_CODECS.md) or [`cosmjs-types`](https://github.com/confio/cosmjs-types/blob/main/scripts/codegen.js) do is a good choice.
+You can choose which library you use to compile your Protobuf objects into TypeScript or JavaScript. Reproducing what [cosmjs-types](https://github.com/confio/cosmjs-types/blob/main/scripts/codegen.js) or [Stargate](https://github.com/cosmos/cosmjs/blob/main/packages/stargate/CUSTOM_PROTOBUF_CODECS.md) do is a good choice.
 
 ### Preparation
 
 This exercise assumes that:
 
-1. Your Protobuf definition files are in `./proto/myChain`.
-2. You want to compile them into TypeScript in `./client/src/types/generated`.
+1. The library you're creating is in 'myLib'.
+2. Your Protobuf definition files are in `./proto/`.
+2. You want to transpile them into TypeScript in `./src/codegen/`.
 
-Install `protoc` on your computer and its Typescript plugin in your project, possibly with the help of a Dockerfile:
-
-<CodeGroup>
-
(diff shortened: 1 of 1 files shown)


tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -31,120 +31,38 @@ This section assumes that you have a working Cosmos blockchain with its own modu
...
-$ ln -s /usr/lib/protoc/bin/protoc /usr/local/bin/protoc
-$ npm install ts-proto@1.121.6 --save-dev
+npm install -g @cosmology/telescope
 ```
 
-Adjust to your preferred version, operating system, and CPU platform. For instance, on an Apple M1 you would use `protoc-21.7-osx-aarch_64.zip`.
-
-</CodeGroupItem>
-
-<CodeGroupItem title="Dockerfile">
```
xavierlepretre, line 63: Please adjust so that we keep both Local and Dockerfile.

tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -31,120 +31,38 @@ This section assumes that you have a working Cosmos blockchain with its own modu
...
 
 ```sh
-$ mkdir -p /usr/lib/protoc
-$ cd /usr/lib/protoc
-$ curl -L https://github.com/protocolbuffers/protobuf/releases/download/v21.7/protoc-21.7-linux-x86_64.zip -o protoc.zip
-$ unzip -o protoc.zip
-$ rm protoc.zip
-$ ln -s /usr/lib/protoc/bin/protoc /usr/local/bin/protoc
-$ npm install ts-proto@1.121.6 --save-dev
+npm install -g @cosmology/telescope
```
xavierlepretre, line 47: We prefer to save locally in the project instead of `-g` globally. 
Is it a `--save-dev` here or not?
Please also set the `@x.y.z` version at which this update works. Also, I think we forgot a `--save-exact` here, if you could add it, so that we maximise reproducibility.
Also, keep the `$` as it is used everywhere in the tutorials I think.

tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -31,120 +31,38 @@ This section assumes that you have a working Cosmos blockchain with its own modu
...
-
-</CodeGroupItem>
-
-<CodeGroupItem title="Docker">
+You can confirm the version you received by running:
 
 ```sh
-$ docker run --rm -it \
-    ts-protoc --version
+$ telescope -v
```
xavierlepretre, line 55: If `--version` exists, please use the long form so that it is more easily understandable by the person skimming the content. Sometimes `-v` means _verbose_.

tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -184,59 +102,88 @@ $ mkdir -p ./proto/gogoproto
...
+? [pkg] which packages do you want to support? (Press <space> to select, <a> to toggle all, <i> to invert selection)
+❯◯ amino
+ ◯ akash
+ ◯ bcna
+ ...
+ ◯ cosmos
+ ...
+```
+
+Say `cosmos`'s selected and entered, there'll be cosmos folder and its dependencies(like gogoproto and google, etc) in `myLib/proto` folder.
```
xavierlepretre, line 124: Is there a non-interactive command? If yes, then perhaps a `CodeGroup` is welcome with "interactive" and "non-interactive".

tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -249,9 +196,9 @@ service Msg {
 }
 ```
 
-If so, you find its service declaration in the compiled `tx.ts` file:
+If so, you find its service declaration in the compiled `tx.ts` file(with telescope option rpcClients.inline: true) or in `tx.rpc.msg.ts` file(with telescope option rpcClients.inline: false):
```
xavierlepretre, line 199:
```suggestion
If so, you find its service declaration in the compiled `tx.ts` file (with telescope option `rpcClients.inline: true`) or in `tx.rpc.msg.ts` file(with telescope option `rpcClients.inline: false`):
```

tutorials/7-cosmjs/5-create-custom.md
```diff
@@ -249,9 +196,9 @@ service Msg {
...
 
-```typescript [https://github.com/confio/cosmjs-types/blob/v0.4.1/src/cosmos/bank/v1beta1/tx.ts#L243-L248]
+```typescript [https://github.com/confio/cosmjs-types/blob/4bce108bc9307dc17f7127d6ff7de80c867c4a09/src/cosmos/bank/v1beta1/tx.ts#L484]
```
xavierlepretre, line 201: If you have tags instead of commits, that's better.

tutorials/7-cosmjs/5-create-custom.md
Zetazzz: Hi, Thank you very much for the review! 
Please excuse me that I don't understand how can I keep the Dockerfile part for this. I think with --save-dev and --save-exact as you suggested, telescope can be installed locally in the package, so we don't have to use docker in this case. Did you mean that I should also keep protoc part? I think that will make this tutorial more complicated... 
Other suggestions are great! I'll fix them in the PR in developer-portal

xavierlepretre: If we don't need the `protoc` executable anymore because Telescope is taking care of this, and it is a single `npm` package, then perhaps we don't need Docker.

xavierlepretre:
It was moved here.https://github.com/cosmos/developer-portal/pull/38

PR #1494, opened 2024-01-04
feat: Oracle tutorial and example

samricotta:


tac0turtle:
pushed changes in accordance with what we talked. when youre back can you review and mark it r4r if its good to go, lets get this merged then bring over the docs

